Issue #7, opened 2022-10-20
gromacs test runs out of memory


ladinesa:
@JFRudzinski I commented out test_radius_of_gyration in gromacs tests because it is cosuming too much memory. Can you please have a look at this?

JFRudzinski:
Yes, I will take a look tomorrow and get back to you.

JFRudzinski:
I checked to make sure nothing strange was going on within the calculation of the radius of gyration. It appears that the memory usage (I saw ~2gb in my profiling) is simply due to the "normal" processing of this somewhat large system. I don't know how you would run out of memory though, I can run the tests perfectly fine on my laptop.

In any case, I will work on constructing a smaller system / shorter simulation for testing purposes and replace the test. 

I will also plan to look a bit closer into the various functions that I implemented and try to make sure that nothing funny is going on. Please leave this issue open for now until I run some more tests.

JFRudzinski:
@ladinesa : Please re-check this memory issue using the branch from Issue #13 . I think this may have fixed the problem.

JFRudzinski:
@ladinesa: Did you confirm that the memory issue is fixed? Can we close this issue now?

ladinesa:
It still takes ca. 2 mins for the test_radius_of_gyration, but at least now my computer does not freeze so we can close this.

JFRudzinski:
hmm, that's strange, it doesn't take so long on my computer, maybe 30s. Well, I can also curate a smaller system for this test at some point. Maybe we should actually leave it open so that I don't forget?

JFRudzinski:
Solved by https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/tree/1490-smaller-test-data-for-md-workflow

PR #14, opened 2022-11-21
13 invalid value encountered in true divide in mda function for fsfg example

### LINKED ISSUE #13: Invalid value encountered in true_divide in MDA function for fsfg example
Problem:
.../MDAnalysis/core/groups.py:1099: RuntimeWarning: invalid value encountered in true_divide
  _centers /= _weights.sum(axis=1)[:, None]

Note:
This may somehow be related to the memory issue in the Rg test which uses the fsfg example.

JFRudzinski:
There was an issue here because the fsfg example (also used for the Rg test function in the gromacs tests) uses a 4-site water model with virtual sites. In this case, there was an issue with how the center of mass is calculated because the virtual sites return NAN after trying to wrap the coordinates. 

Solution: I check for massless sites and remove them from the definition of COM beads in the MDA utility.

The fixes here may solve the memory issue with the Rg test - Issue #7


atomisticparsers/utils/mdanalysis.py
```diff
@@ -70,6 +70,24 @@ def universe(self):
...
 
+    @property
+    def bead_groups(self):
+        atoms_moltypes = self.get('atoms_info', {}).get('moltypes', [])
+        moltypes = np.unique(atoms_moltypes)
+        bead_groups = {}
+        compound = 'fragments'
+        for moltype in moltypes:
+            if hasattr(self.universe.atoms, 'moltypes'):
+                AGs_by_moltype = self.universe.select_atoms('moltype ' + moltype)
```
ladinesa: use small case

atomisticparsers/utils/mdanalysis.py
```diff
@@ -193,69 +211,74 @@ def get_rdf_avg(rdf_results_tmp, rdf_results, interval_indices, n_frames_split):
...
-                    rdf_results_tmp['bins'].append(rdf.results.bins[int(n_smooth / 2):-int(n_smooth / 2)] * ureg.angstrom)
-                    rdf_results_tmp['value'].append(np.convolve(
-                        rdf.results.rdf, np.ones((n_smooth,)) / n_smooth,
-                        mode='same')[int(n_smooth / 2):-int(n_smooth / 2)])
-
-                for interval_group in interval_indices:
-                    get_rdf_avg(rdf_results_tmp, rdf_results, interval_group, n_frames_split)
-
-        return rdf_results
+        # bead_groups = {}
```
ladinesa: remove commented. either they are there or not.

atomisticparsers/utils/mdanalysis.py
```diff
@@ -70,6 +70,24 @@ def universe(self):
                 self.logger.error('Error creating MDAnalysis universe.', exc_info=e)
         return self._file_handler
 
+    @property
```
ladinesa, line 73: I think property is not appropriate here. Normally you use property when you have a quantity that you use multiple times should not be recalculated. Maybe just name it `calculate_bead_groups`?

atomisticparsers/utils/mdanalysis.py
```diff
@@ -193,69 +211,62 @@ def get_rdf_avg(rdf_results_tmp, rdf_results, interval_indices, n_frames_split):
...
-                        rdf.results.rdf, np.ones((n_smooth,)) / n_smooth,
-                        mode='same')[int(n_smooth / 2):-int(n_smooth / 2)])
-
-                for interval_group in interval_indices:
-                    get_rdf_avg(rdf_results_tmp, rdf_results, interval_group, n_frames_split)
-
-        return rdf_results
+        if bead_groups is None:
+            return {}
+        else:
```
ladinesa: You do not need the else here as you already return when bead_groups is None

atomisticparsers/utils/mdanalysis.py
JFRudzinski: I made it a property because it is something that (so far) is needed in calculate_molecular_rdf() and calculate_molecular_mean_squared_displacement(). You may access neither, one, or both of these. So, I wanted to ensure that you only calculated the bead_groups once if necessary.

ladinesa:
Please mark resolve when issue is resolved so I can approve it.

atomisticparsers/utils/mdanalysis.py
JFRudzinski: done

atomisticparsers/utils/mdanalysis.py
JFRudzinski: done

atomisticparsers/utils/mdanalysis.py
JFRudzinski: done

ladinesa (approved):

ladinesa:
Please fix linting error before merging.

PR #22, opened 2022-12-15
19 generalization of md workflow results

### LINKED ISSUE #19: Generalization of MD workflow results
The workflow results (e.g., rdf/msd) are currently calculated in each atomistic parser from the MDA instances created from the corresponding raw files. These calculations should be moved to the normalization step and be computed from a (generalized) MDA instance created directly from the nomad metainfo.
nb - as a side effect, we should store the references to each molecule / molecule group for these observables using the topology structure with the results section.

(NOMAD issue [#1199](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/issues/1199))

JFRudzinski:


JFRudzinski:
@ladinesa : So, this is the updated parser code that goes with the changes that we just merged to nomad dev. There are a bunch of import issues though, do you know what is happening? Is it not linked to the correct nomad version?

ladinesa:
this is expected as the target nomad in dependencies/parsers/atomistic/.github/workflows/python-actions.yaml is v1.1.1 which does not still contain the metainfo defs in nomad develop. To remedy this, one has to release a pypi package with develop which will only happen when we decide to release a new nomad version hopefully in the next months. For now, we simply have to run the tests locally and make sure they all pass before pushing changes.

JFRudzinski:
oh, ok, I see. Well, all my normal tests work. Do you want to review the code when you have a chance? Mostly it is just removing some code, since I moved things outside to the nomad atomutils and workflow normalizers.

ladinesa:
I have quicky looked at the changes and they seem to be reasonable, also containing the changes i did recently to include workflow2. Please merge. I think the next step is to then to replace workflow with workflow2. As done with the normalization, I will start in nomad then you continue with atomistic parsers.

PR #31, opened 2023-01-18
28 connect system and calculation when non matching lengths

### LINKED ISSUE #28: Connect system and calculation when non-matching lengths
We need to correctly reference between the systems and calculations in the case that they are differing in lengths. 

Once this is done, the calculation of Rg in the workflow normalizer of NOMAD should be updated to account for this.

JFRudzinski:
@ladinesa - I have re-written the parse_thermodynamic_data() function of both parsers to match the calculation and system lists through "system_ref". The approach is to document the time and step within both lists and then check for a match when parsing the calculation list. I am sure that you will be able to improve my code significantly, I was unsure about quite a number of things!

nb - connected to branch 1289-connect-system-and-calculation-when-non-matching-lengths of nomad

### DIFF
--- a/atomisticparsers/gromacs/parser.py
+++ b/atomisticparsers/gromacs/parser.py
@@ -728,7 +728,8 @@ class GromacsParser:
             sec_system.time = self.traj_parser.get_time(n)  # TODO Physical times should not be stored for GeometryOpt
             time_step = self.log_parser.get('input_parameters', {}).get('dt', 1.0) * ureg.ps
             if sec_system.time is not None:
-                sec_system.step = int(sec_system.time / time_step) if time_step else None
+                sec_system.step = int(ureg.convert(sec_system.time.magnitude, sec_system.time.units, time_step.units) / time_step.magnitude) if time_step else None
+                # sec_system.step = int(sec_system.time / time_step) if time_step else None
             if positions is None:
                 continue
 
--- a/atomisticparsers/gromacs/parser.py
+++ b/atomisticparsers/gromacs/parser.py
@@ -616,7 +616,6 @@ class GromacsParser:
 
     def parse_thermodynamic_data(self):
         sec_run = self.archive.run[-1]
-        # sec_system = sec_run.system
 
         n_frames = self.traj_parser.get('n_frames')
         time_step = self.log_parser.get('input_parameters', {}).get('dt', 1.0) * ureg.ps
--- a/atomisticparsers/gromacs/parser.py
+++ b/atomisticparsers/gromacs/parser.py
@@ -619,6 +619,7 @@ class GromacsParser:
         sec_system = sec_run.system
 
         n_frames = self.traj_parser.get('n_frames')
+        time_step = self.log_parser.get('input_parameters', {}).get('dt', 1.0) * ureg.ps
 
         # get it from edr file
         if self.energy_parser.keys():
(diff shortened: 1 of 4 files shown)


JFRudzinski:
@ladinesa  - review reminder

atomisticparsers/gromacs/parser.py
```diff
@@ -712,6 +725,11 @@ def get_composition(children_names):
                 continue
             positions = self.traj_parser.get_positions(n)
             sec_system = sec_run.m_create(System)
+            sec_system.time = self.traj_parser.get_time(n)  # TODO Physical times should not be stored for GeometryOpt
```
ladinesa, line 711: we can probably have a mapping here, from system index to time. 
```
system_time_map = {round(self.traj_parser.get_time(n), 5): n for n in range(len(sec_system))}
```
We round it so we do not need to calculate diff but simply also round the calc time during query.

atomisticparsers/lammps/parser.py
```diff
@@ -344,6 +344,12 @@ def get_n_atoms(self, idx):
             return len(self.get_positions(idx))
         return n_atoms[idx]
 
+    def get_step(self, idx):
```
ladinesa, line 347: I think we do not need this function. simply use `self.get('time_step', idx)`

atomisticparsers/gromacs/parser.py
```diff
@@ -652,25 +647,43 @@ def parse_thermodynamic_data(self):
...
+            1.0, calculation_times_ps.units, ureg.picosecond)
+
+        map_calculation_to_system = {}
+        for i_calc, calc_time in enumerate(calculation_times_ps):
+            if calc_time is None:
+                calc_index = system_steps.index(calculation_steps[i_calc]) if calculation_steps[i_calc] in system_steps else None
+                map_calculation_to_system[str(i_calc)] = calc_index
+                continue
+            flag_match = False
+            for i_sys, sys_time in enumerate(system_times_ps):
```
ladinesa: I do not like the double loop here, I think we can fix this using the system_time_map

atomisticparsers/gromacs/parser.py
```diff
@@ -652,25 +647,43 @@ def parse_thermodynamic_data(self):
...
-            create_scc = True
+        system_steps = [frame.step for frame in sec_system]
+        system_times_ps = [frame.time.magnitude * ureg.convert(
+            1.0, frame.time.units, ureg.picosecond) for frame in sec_system]
+        calculation_times_ps = thermo_data.get('Time')
+        calculation_steps = [int((val / time_step).magnitude) for val in calculation_times_ps]
+        calculation_times_ps = calculation_times_ps.magnitude * ureg.convert(
+            1.0, calculation_times_ps.units, ureg.picosecond)
+
+        map_calculation_to_system = {}
```
ladinesa: we do not need this map i guess, we can directly create calculation sections here?

atomisticparsers/lammps/parser.py
```diff
@@ -648,33 +654,52 @@ def get_time_step(self):
...
-
         n_thermo = len(thermo_data.get('Step', []))
-        create_scc = True
-        if sec_sccs:
-            if len(sec_sccs) != n_thermo:
-                self.logger.warning(
-                    '''Mismatch in number of calculations and number of property
-                    evaluations!, will create new sections''',
-                    data=dict(n_calculations=len(sec_sccs), n_evaluations=n_thermo))
+        system_steps = [frame.step for frame in sec_system]
```
ladinesa: My comments are the same as for gromacs. We can perhaps make a super class for MD parsers. which has only this parse_thermodynamic data so we do not write two codes. This will also include the frame sampling rate. But we can do it in a separate issue.

atomisticparsers/gromacs/parser.py
```diff
@@ -564,7 +564,7 @@ def __init__(self):
             'Potential': 'potential', 'Kinetic En.': 'kinetic'}
         self._frame_rate = None
         # max cumulative number of atoms for all parsed trajectories to calculate sampling rate
-        self._cum_max_atoms = 1000000
```
ladinesa, line 567: maybe crreate an issue where we create a super class

atomisticparsers/gromacs/parser.py
JFRudzinski: done (I kept it inside the system loop though)

atomisticparsers/gromacs/parser.py
JFRudzinski: Thank you for the fix. I knew this was bad but I wasn't sure how to fix it. done.

atomisticparsers/gromacs/parser.py
JFRudzinski: removed.

atomisticparsers/gromacs/parser.py
ladinesa: I would have done the same on the first try.

atomisticparsers/lammps/parser.py
JFRudzinski: I opened an issue for this.

JFRudzinski:
@ladinesa - I think it is much better now, but please re-check my changes :pray:

atomisticparsers/gromacs/parser.py
```diff
@@ -616,15 +616,10 @@ def get_gromacs_file(self, ext):
 
     def parse_thermodynamic_data(self):
         sec_run = self.archive.run[-1]
+        # sec_system = sec_run.system
```
ladinesa: please remove this.

ladinesa:
Great thanks for incorporating my suggestions. You can merge it now.

PR #34, opened 2023-01-24
added system counter to deal with pruning within the time map

### LINKED ISSUE #33: system time map error when frame rate not 1

JFRudzinski:
@ladinesa - I had to add a counter within the system loop in parse_system for the case that we are pruning the trajectory with frame_rate, so that the system_time_map is correct.

### DIFF
--- a/atomisticparsers/gromacs/parser.py
+++ b/atomisticparsers/gromacs/parser.py
@@ -703,6 +703,7 @@ class GromacsParser:
         pbc = self.log_parser.get_pbc()
         self._system_time_map = {}
         self._system_step_map = {}
+        system_counter = 0
         for n in range(n_frames):
             if (n % self.frame_rate) > 0:
                 continue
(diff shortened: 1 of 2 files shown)


atomisticparsers/gromacs/parser.py
```diff
@@ -712,12 +713,12 @@ def get_composition(children_names):
             time_step = self.log_parser.get('input_parameters', {}).get('dt', 1.0) * ureg.ps
             if sec_system.time is not None:
                 self._system_time_map[round(ureg.convert(
-                    sec_system.time.magnitude, sec_system.time.units, ureg.picosecond), 5)] = n
+                    sec_system.time.magnitude, sec_system.time.units, ureg.picosecond), 5)] = system_counter
```
ladinesa, line 716: I think we do not need system_counter, simply len(system_step_map).

ladinesa:
Oh yes, i missed this. Please merge.

atomisticparsers/gromacs/parser.py
ladinesa: please include this in next pr

JFRudzinski: oh sorry, I misunderstood, I thought you were saying nevermind below. I will change it now.

ladinesa: It's ok its minor, can go with another pr.

PR #39, opened 2023-01-27
created calculation entries for all system entries. create calc/sys r…

### LINKED ISSUE #38: create calculation entries for each item in the system list

JFRudzinski:
…eference for equal lengths in case time/step info not available

Calculation list now contains entries for every member of system, regardless if corresponding calculation info exists. The matching is the same as before (based on time or step info). The final calculation list is sorted by time or step info if available.

In case there is no time or step info available, there are two possible cases:
1. system and calculation have the same length: then the references are created assuming correspondence and a warning is given.
2. system and calculation have different lengths: then the missing system entries are appended to the calculation list and a warning is given that the list may not be in order.

### DIFF
--- a/atomisticparsers/gromacs/parser.py
+++ b/atomisticparsers/gromacs/parser.py
@@ -616,6 +616,7 @@ class GromacsParser:
 
     def parse_thermodynamic_data(self):
         sec_run = self.archive.run[-1]
+        sec_system = sec_run.system
 
         n_frames = self.traj_parser.get('n_frames')
         time_step = self.log_parser.get('input_parameters', {}).get('dt', 1.0) * ureg.ps
@@ -651,6 +652,7 @@ class GromacsParser:
         calculation_times_ps = calculation_times_ps.magnitude * ureg.convert(
             1.0, calculation_times_ps.units, ureg.picosecond)
 
+        system_indices_assigned = []
         for n in range(n_evaluations):
             sec_scc = sec_run.m_create(Calculation)
             system_index = self._system_time_map.get(
@@ -658,8 +660,9 @@ class GromacsParser:
             if system_index is None:
                 system_index = self._system_step_map.get(calculation_steps[n])
             if system_index is not None:
+                system_indices_assigned.append(system_index)
                 sec_scc.forces = Forces(total=ForcesEntry(value=self.traj_parser.get_forces(system_index)))
-                sec_scc.system_ref = sec_run.system[system_index]
+                sec_scc.system_ref = sec_system[system_index]
                 sec_scc.method_ref = sec_run.method[-1] if sec_run.method else None
 
             # TODO add other energy contributions, properties
@@ -690,6 +693,33 @@ class GromacsParser:
                     sec_energy.contributions.append(
                         EnergyEntry(kind=self._metainfo_mapping[key], value=val))
 
+        if not self._system_step_map and not self._system_time_map:
+            if n_evaluations == len(sec_system):
+                for i_calc, calc in enumerate(sec_run.calculation):
+                    calc.system_ref = sec_system[i_calc]
+                self.logger.warning('Time and step information not available, but system and calculation are the same length.'
+                                    'Assuming correspondence and creating references between these two lists.')
+                return
+
+        for n in range(len(sec_system)):
+            if n not in system_indices_assigned:
+                sec_scc = sec_run.m_create(Calculation)
+                sec_scc.time = self._system_info[n]['time']
+                sec_scc.step = self._system_info[n]['step']
+
+        times = [calc.time for calc in sec_run.calculation]
+        steps = [calc.step for calc in sec_run.calculation]
+        if None not in times:
+            calculations_sorted = [[calc.time.magnitude, calc] for calc in sec_run.calculation]
+            calculations_sorted = sorted(calculations_sorted, key=lambda x: x[0], reverse=False)
+            sec_run.calculation = [calc[1] for calc in calculations_sorted]
+        elif None not in steps:
+            calculations_sorted = [[calc.step, calc] for calc in sec_run.calculation]
+            calculations_sorted = sorted(calculations_sorted, key=lambda x: x[0], reverse=False)
+            sec_run.calculation = [calc[1] for calc in calculations_sorted]
+        else:
+            self.logger.warning('Time and step information not available in section calculation, entries may be out of order.')
+
     def parse_system(self):
         sec_run = self.archive.run[-1]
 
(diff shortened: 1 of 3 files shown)


atomisticparsers/gromacs/parser.py
```diff
@@ -651,15 +652,17 @@ def parse_thermodynamic_data(self):
         calculation_times_ps = calculation_times_ps.magnitude * ureg.convert(
             1.0, calculation_times_ps.units, ureg.picosecond)
 
+        system_indices_assigned = []
         for n in range(n_evaluations):
```
ladinesa: We can simplify this by looping over all sec_system. We convert calculation_times_ps to a mapping to get the thermo_data which we assign to the calculation. With this, we do not need line 696-722.

atomisticparsers/gromacs/parser.py
```diff
@@ -703,21 +733,23 @@ def get_composition(children_names):
         pbc = self.log_parser.get_pbc()
         self._system_time_map = {}
         self._system_step_map = {}
+        self._system_info = []
```
ladinesa: We only need system_time_map if we implement the calculation_time_step mapping.

atomisticparsers/lammps/parser.py
```diff
@@ -672,15 +672,17 @@ def parse_thermodynamic_data(self):
             val.magnitude * ureg.convert(1.0, val.units, ureg.picosecond)
             for val in calculation_times_ps] if type(time_step) != float else [None] * len(calculation_times_ps)
 
+        system_indices_assigned = []
```
ladinesa: we can apply the same approach here.

atomisticparsers/gromacs/parser.py
JFRudzinski: If I understand correctly, your proposal does not take into account the situation when there are entries in the calculation section that do not correspond to a system. I still need to create a calculation entry there. Since the list in calculation and the list in system do not have to overlap at all, I don't see any way to avoid going through both lists here. If you still disagree, then I am missing something...?

ladinesa: We can pop the values in thermo data map, if there are still time steps left, then we create additional calculation sections in another loop. The question only is the sorting, but this can be done in a couple of lines, i guess. this themo_data_map should only be a mapping from time to themo_data index

ladinesa: I think my idea is stupid sorry. So we loop over each thermo_data just the same and assign system_ref if time matched and pop the value from the system_time_map. If there are entries left in system_time_map, we loop over them and create addtional calculation sections.

JFRudzinski: Is there a better way to keep track of the remaining entries in system_time_map, so that we don't need to loop over the full system list? 

I thought now of making an array of the system indices and then deleting the ones used in the thermo loop, but probably there is a better way.

ladinesa: popping it like system_time.pop(time, None) will pop it if it exists other will return None.

JFRudzinski: aha! I had no idea what you were referring to when you said "popping" before, which really added to my confusion. I did not know about this function. Ok, I will adjust the code, thanks!

JFRudzinski:
@ladinesa - ok, take another look at these changes. I think it is better now.

atomisticparsers/gromacs/parser.py
```diff
@@ -690,6 +689,27 @@ def parse_thermodynamic_data(self):
...
+                self.logger.warning('Time and step information not available, but system and calculation are the same length.'
+                                    'Assuming correspondence and creating references between these two lists.')
+                return
+
+        for time, n in self._system_time_map.items():
+            sec_scc = sec_run.m_create(Calculation)
+            sec_scc.time = time
+            sec_scc.step = int((time / time_step).magnitude)
+
+        times = [calc.time for calc in sec_run.calculation]
```
ladinesa: use calculation_times_ps, is it the same?

atomisticparsers/gromacs/parser.py
```diff
@@ -690,6 +689,27 @@ def parse_thermodynamic_data(self):
                     sec_energy.contributions.append(
                         EnergyEntry(kind=self._metainfo_mapping[key], value=val))
 
+        if not flag_system_time_map:
```
ladinesa: you do not need  this, simply use if not system_time_map. but do we still need this? is everything not yet covered by line 661?

atomisticparsers/gromacs/parser.py
```diff
@@ -690,6 +689,27 @@ def parse_thermodynamic_data(self):
...
+                                    'Assuming correspondence and creating references between these two lists.')
+                return
+
+        for time, n in self._system_time_map.items():
+            sec_scc = sec_run.m_create(Calculation)
+            sec_scc.time = time
+            sec_scc.step = int((time / time_step).magnitude)
+
+        times = [calc.time for calc in sec_run.calculation]
+        if None not in times:
```
ladinesa: I do like this sorting, maybe we can get rid of this if we concatenate all times from calculation and system and loop over this.

atomisticparsers/gromacs/parser.py
ladinesa: maybe we can still improve this if we concatenate all the times from calculation and system.

atomisticparsers/gromacs/parser.py
JFRudzinski: no, it's not the same. times is now the full list of (non-overlapping) times from the thermo list (which is the same as calculation_times_ps) and the system list, if times are available. But perhaps there is still a shortcut that I am not seeing?

atomisticparsers/gromacs/parser.py
JFRudzinski: 1. The reason I set this flag is that self._system_time_map is changed by popping values in the above thermo loop. I want to know if self._system_time_map was empty at the beginning, not at the end of the popping.

2. I'm not sure if this ever happens for Gromacs, but for Lammps there are cases where the time/step information is completely unavailable, however, the calculation and system lists are the same length. So, I think we want to take the risk in this case and assume that they are corresponding (this is what we were doing before, such that if you don't have this some of the Lammps tests fail).

atomisticparsers/gromacs/parser.py
JFRudzinski: Definitely open to suggestions here, but not exactly sure what you mean.

`times` is the the concatenation of all times from calculation and system, already removing overlapping times...the calculation times come first and then the remaining system times are at the end. I am not sure how to avoid sorting here though, since the (non-)overlap of these two lists is completely arbitrary...?

JFRudzinski:
@ladinesa - ok, let's see what you think about this now. 

Note that I no longer take care of this case in lammps where times are not available in system. Lammps always has step info and gromacs always has time info (to my knowledge), so the mappings are done with times and steps, respectively, in each parser.

PR #40, opened 2023-03-07
Upgrade python version

markus1978:
This [MR](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1125) on nomad-fair (gitlab), which updates the minimum required Python to 3.8 and will run the official nomand and all images on Python 3.9, requires these changes.


ladinesa:
@markus1978 should the target be develop

markus1978:
I am not even sure yet if this is strictly necessary. Also the branches are a bit confusing, because develop is not what the current develop in nomad-fair is using?

This is why we have so many commits. Only one is from me.

PR #48, opened 2023-06-05
fixed units for time in calculation. Checked time for pruned traj fil…

### LINKED ISSUE #47: Calculation times are incorrect for gromacs data
Just noticed that calculation times are on the order of seconds for all gromacs data.

Also need to support the situation that someone prunes the original trajectory file to upload to NOMAD. Then there will be an inconsistency with the expected length. But the times in the trajectory file can be assumed to be correct.

JFRudzinski:
…es, ok!

Just a tiny fix. units were accidentally removed when making time map between system and calculation.

### DIFF
--- a/atomisticparsers/gromacs/parser.py
+++ b/atomisticparsers/gromacs/parser.py
@@ -658,7 +658,7 @@ class GromacsParser:
 
         for time in sorted(time_map):
             sec_scc = sec_run.m_create(Calculation)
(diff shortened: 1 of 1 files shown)


JFRudzinski:
@ladinesa - Just a tiny fix. units were accidentally removed when making time map between system and calculation. Still, just give me the ok to merge please.

ladinesa (approved):

ladinesa:
please merge and close this.

PR #57, opened 2023-08-03
56 problems with gromacs parser from umbrella sampling simulation

### LINKED ISSUE #56: Problems with Gromacs parser from umbrella sampling simulation
A user posted some issues with the Gromacs parser for a pair of umbrella sampling simulations. It is unclear what the issue is upon first glance (probably there are several issues). Will debug here...

JFRudzinski:
@ladinesa I think this is done. Again the main issues with the users data that I found were:
1. regex for Energy block
2. version of MDAnalysis
3. restrictions for calculating rdf and msd for too many molecules

Please review my code and let me know what you think....Thanks!

### DIFF
--- a/atomisticparsers/gromacs/metainfo/gromacs.py
+++ b/atomisticparsers/gromacs/metainfo/gromacs.py
@@ -20,7 +20,7 @@ import numpy as np            # pylint: disable=unused-import
 import typing                 # pylint: disable=unused-import
 from nomad.metainfo import (  # pylint: disable=unused-import
     MSection, MCategory, Category, Package, Quantity, Section, SubSection, SectionProxy,
-    Reference
+    Reference, JSON
 )
 from nomad.datamodel.metainfo import simulation
 
@@ -2041,6 +2041,35 @@ class Entry(MSection):
         ''')
 
 
+class CalcEntry(MSection):
+    '''
+    Section describing a general type of calculation.
+    '''
+
+    m_def = Section(validate=False)
+
+    kind = Quantity(
+        type=str,
+        shape=[],
+        description='''
+        Kind of the quantity.
+        ''')
+
+    value = Quantity(
+        type=np.dtype(np.float64),
+        shape=[],
+        description='''
+        Value of this contribution.
+        ''')
+
+    unit = Quantity(
+        type=str,
+        shape=[],
+        description='''
+        Unit of the parameter as a string.
+        ''')
+
+
 class Calculation(simulation.calculation.Calculation):
 
     m_def = Section(validate=False, extends_base_section=True,)
@@ -2050,11 +2079,16 @@ class Calculation(simulation.calculation.Calculation):
         repeats=True,)
 
     x_gromacs_thermo_contributions = SubSection(
-        sub_section=Entry.m_def,
+        sub_section=CalcEntry.m_def,
         description='''
         Contains other gromacs-specific thermodynamic and energy contributions that are not already defined.
         ''',
         repeats=True)
+
+    x_gromacs_thermo_contributions_json = Quantity(
+        type=JSON,
+        description='''Contains other gromacs-specific thermodynamic and energy contributions that are not already defined.
+        ''')
 
 
 class Energy(simulation.calculation.Energy):
@@ -2067,3 +2101,8 @@ class Energy(simulation.calculation.Energy):
         Contains other gromacs-specific energy contributions that are not already defined.
         ''',
         repeats=True)
+
+    x_gromacs_energy_contributions_json = Quantity(
+        type=JSON,
+        description='''Contains other gromacs-specific thermodynamic and energy contributions that are not already defined.
+        ''')
--- a/atomisticparsers/gromacs/metainfo/gromacs.py
+++ b/atomisticparsers/gromacs/metainfo/gromacs.py
@@ -2055,3 +2055,15 @@ class Calculation(simulation.calculation.Calculation):
         Contains other gromacs-specific thermodynamic and energy contributions that are not already defined.
         ''',
         repeats=True)
+
+
+class Energy(simulation.calculation.Energy):
+
+    m_def = Section(validate=False, extends_base_section=True,)
+
+    x_gromacs_energy_contributions = SubSection(
+        sub_section=simulation.calculation.EnergyEntry.m_def,
+        description='''
+        Contains other gromacs-specific energy contributions that are not already defined.
+        ''',
+        repeats=True)
--- a/atomisticparsers/gromacs/metainfo/gromacs.py
+++ b/atomisticparsers/gromacs/metainfo/gromacs.py
@@ -2063,7 +2063,6 @@ class Calculation(simulation.calculation.Calculation):
         repeats=True)
 
 
-
 class Energy(simulation.calculation.Energy):
 
     m_def = Section(validate=False, extends_base_section=True,)
@@ -2074,4 +2073,3 @@ class Energy(simulation.calculation.Energy):
         Contains other gromacs-specific energy contributions that are not already defined.
         ''',
         repeats=True)
-
--- a/atomisticparsers/gromacs/metainfo/gromacs.py
+++ b/atomisticparsers/gromacs/metainfo/gromacs.py
@@ -2019,6 +2019,28 @@ class Constraint(simulation.system.Constraint):
         repeats=True,)
 
 
+class Entry(MSection):
+    '''
+    Section describing a type of energy or a contribution to the total energy.
+    '''
+
+    m_def = Section(validate=False)
+
+    kind = Quantity(
+        type=str,
+        shape=[],
+        description='''
+        Kind of the quantity.
+        ''')
+
+    value = Quantity(
+        type=np.dtype(np.float64),
+        shape=[],
+        description='''
+        Value of this contribution (units implied by Gromacs unit defaults).
+        ''')
+
+
 class Calculation(simulation.calculation.Calculation):
 
     m_def = Section(validate=False, extends_base_section=True,)
@@ -2026,3 +2048,10 @@ class Calculation(simulation.calculation.Calculation):
     x_gromacs_section_single_configuration_calculation = SubSection(
         sub_section=SectionProxy('x_gromacs_section_single_configuration_calculation'),
         repeats=True,)
+
+    x_gromacs_thermo_contributions = SubSection(
+        sub_section=Entry.m_def,
+        description='''
+        Contains other gromacs-specific thermodynamic and energy contributions that are not already defined.
+        ''',
+        repeats=True)
--- a/atomisticparsers/gromacs/metainfo/gromacs.py
+++ b/atomisticparsers/gromacs/metainfo/gromacs.py
@@ -19,8 +19,7 @@
 import numpy as np            # pylint: disable=unused-import
 import typing                 # pylint: disable=unused-import
 from nomad.metainfo import (  # pylint: disable=unused-import
-    MSection, MCategory, Category, Package, Quantity, Section, SubSection, SectionProxy,
-    Reference, JSON
+    MSection, MCategory, Category, Package, Quantity, Section, SubSection, SectionProxy
 )
 from nomad.datamodel.metainfo import simulation
 
@@ -2019,28 +2018,6 @@ class Constraint(simulation.system.Constraint):
         repeats=True,)
 
 
-class Entry(MSection):
-    '''
-    Section describing a type of energy or a contribution to the total energy.
-    '''
-
-    m_def = Section(validate=False)
-
-    kind = Quantity(
-        type=str,
-        shape=[],
-        description='''
-        Kind of the quantity.
-        ''')
-
-    value = Quantity(
-        type=np.dtype(np.float64),
-        shape=[],
-        description='''
-        Value of this contribution (units implied by Gromacs unit defaults).
-        ''')
-
-
 class CalcEntry(MSection):
     '''
     Section describing a general type of calculation.
@@ -2078,17 +2055,13 @@ class Calculation(simulation.calculation.Calculation):
         sub_section=SectionProxy('x_gromacs_section_single_configuration_calculation'),
         repeats=True,)
 
-    x_gromacs_thermo_contributions = SubSection(
+    x_gromacs_thermodynamics_contributions = SubSection(
         sub_section=CalcEntry.m_def,
         description='''
         Contains other gromacs-specific thermodynamic and energy contributions that are not already defined.
         ''',
         repeats=True)
 
-    x_gromacs_thermo_contributions_json = Quantity(
-        type=JSON,
-        description='''Contains other gromacs-specific thermodynamic and energy contributions that are not already defined.
-        ''')
 
 
 class Energy(simulation.calculation.Energy):
@@ -2102,7 +2075,3 @@ class Energy(simulation.calculation.Energy):
         ''',
         repeats=True)
 
-    x_gromacs_energy_contributions_json = Quantity(
-        type=JSON,
-        description='''Contains other gromacs-specific thermodynamic and energy contributions that are not already defined.
-        ''')
--- a/atomisticparsers/gromacs/parser.py
+++ b/atomisticparsers/gromacs/parser.py
@@ -545,9 +545,9 @@ class GromacsParser:
         self._gro_energy_units = ureg.kilojoule * MOL
         self._thermo_ignore_list = ['Time', 'Box-X', 'Box-Y', 'Box-Z']
         self._tensor_index_map = {
-            'XX': np.array([0, 0]), 'XY': np.array([0, 1]), 'XZ': np.array([0, 2]),
-            'YX': np.array([1, 0]), 'YY': np.array([1, 1]), 'YZ': np.array([1, 2]),
-            'ZX': np.array([2, 0]), 'ZY': np.array([2, 1]), 'ZZ': np.array([2, 2])}
+            'XX': (0, 0), 'XY': (0, 1), 'XZ': (0, 2),
+            'YX': (1, 0), 'YY': (1, 1), 'YZ': (1, 2),
+            'ZX': (2, 0), 'ZY': (2, 1), 'ZZ': (2, 2)}
         self._base_calc_map = {
             'Temperature': ('temperature', ureg.kelvin),
             'Volume': ('volume', ureg.nm**3),
@@ -692,12 +692,12 @@ class GromacsParser:
                         dir = key.split('-')[1]
                         ind = self._tensor_index_map.get(dir)
                         if ind is not None:
-                            pressure_tensor[ind] = val * ureg.bar
+                            pressure_tensor[ind] = val
                     elif key.startswith('Vir-'):
                         dir = key.split('-')[1]
                         ind = self._tensor_index_map.get(dir)
                         if ind is not None:
-                            virial_tensor[ind] = val * (ureg.bar * ureg.nm**3)
+                            virial_tensor[ind] = val
                     # well-defined, single Energy quantities
                     elif key in self._energy_map.keys():
                         sec_energy.m_add_sub_section(getattr(Energy, self._energy_map[key]), EnergyEntry(value=val * self._gro_energy_units))
@@ -716,8 +716,8 @@ class GromacsParser:
                             sec_scc.x_gromacs_thermo_contributions.append(
                                 Entry(kind=key, value=val))
 
-                sec_scc.pressure_tensor = pressure_tensor
-                sec_scc.virial_tensor = virial_tensor if virial_tensor is not None else None
+                sec_scc.pressure_tensor = pressure_tensor * ureg.bar
+                sec_scc.virial_tensor = virial_tensor * (ureg.bar * ureg.nm**3)
                 if vdw_dict:
                     total = sum(val for _, val in vdw_dict.items())
                     sec_energy.van_der_waals = EnergyEntry(value=total, short_range=vdw_dict.get('short_range'),
--- a/atomisticparsers/gromacs/parser.py
+++ b/atomisticparsers/gromacs/parser.py
@@ -83,28 +83,46 @@ class GromacsLogParser(TextParser):
                     parameters[val_array.group(1)].append(value[0] if len(value) == 1 else value)
             return parameters
 
-        def str_to_energies(val_in):
-            energy_keys_re = re.compile(r'(.+?)(?:  |\Z| (?=P))')
-            keys = []
-            values = []
-            energies = dict()
-            for val in val_in.strip().split('\n'):
-                val = val.strip()
-                if val[0].isalpha():
-                    keys = [k.strip() for k in energy_keys_re.findall(val)]
-                    keys = ['P%s' % k if k.startswith('res') else k for k in keys if k]
-                else:
-                    values = val.split()
-                    print(keys)
-                    for n, key in enumerate(keys):
+        def str_to_energies(energies_block):
 
-                        if key == 'Temperature':
-                            energies[key] = float(values[n]) * ureg.kelvin
-                        elif key.startswith('Pres'):
-                            key = key.rstrip(' (bar)')
-                            energies[key] = float(values[n]) * ureg.bar
-                        else:
-                            energies[key] = float(values[n]) / MOL * ureg.kJ
+            max_char_name = 14
+            rows = energies_block.strip().split("\n")
+            names = [rows[i_row] for i_row in range(len(rows)) if i_row % 2 == 0]
+            values = [rows[i_row] for i_row in range(len(rows)) if i_row % 2 != 0]
+
+            names_re = r"(?<!\S)[A-Za-z0-9.()\/-]{1,}(?: [\S][A-Za-z0-9.()\/-]{1,})*(?!\S)"
+            values_re = r"[-+]?\d+\.\d+e[+-]?\d+"
+            names = [re.findall(names_re, name) for name in names]
+            names = [name for group in names for name in group]
+            names_split = []
+            for name in names:
+                if len(name) <= max_char_name:
+                    names_split.append(name.strip())
+                else:
+                    remainder_len = len(name) % max_char_name
+                    divisor = int(len(name) / max_char_name)
+                    names_split.append(name[:remainder_len - divisor].strip())
+                    for i in range(int(len(name) / max_char_name)):
+                        start = remainder_len - divisor + (i + 1) + i * max_char_name
+                        end = start + max_char_name
+                        names_split.append(name[start:end].strip())
+
+            values = [re.findall(values_re, val) for val in values]
+            values = [val for group in values for val in group]
+
+            if len(names_split) != len(values):
+                self.logger.error('Error matching energy values.')
+
+            energies = {}
+            for key, val in zip(names_split, values):
+                if key == 'Temperature':
+                    energies[key] = float(val) * ureg.kelvin
+                elif key.startswith('Pres'):
+                    key = key.rstrip(' (bar)')
+                    energies[key] = float(val) * ureg.bar
+                else:
+                    energies[key] = float(val) / MOL * ureg.kJ
+
             return energies
 
         def str_to_step_info(val_in):
@@ -116,11 +134,11 @@ class GromacsLogParser(TextParser):
         thermo_quantities = [
             Quantity(
                 'energies',
-                r'Energies \(kJ/mol\)\s*([\s\S]+?)\n\n',
+                r'Energies \(kJ/mol\)\s*([\s\S]+?)\nD.* step.* load imb.*|\n\n',
                 str_operation=str_to_energies, convert=False),
             Quantity(
                 'step_info',
-                r'(Step.+\n[\d\.\- ]+)',
+                r'[\r\n]\s*(Step.+\n[\d\.\- ]+)',
                 str_operation=str_to_step_info, convert=False)]
 
         self._quantities = [
@@ -132,9 +150,9 @@ class GromacsLogParser(TextParser):
             Quantity('execution_path', r'Executable:\s*(.+)'),
             Quantity('working_path', r'Data prefix:\s*(.+)'),
             # TODO cannot understand treatment of the command line in the old parser
-            # Quantity(
-            #     'header',
-            #     r'(?:GROMACS|Gromacs) (2019[\s\S]+?)\n\n', str_operation=str_to_header),
+            Quantity(
+                'header',
+                r'(?:GROMACS|Gromacs) (2019[\s\S]+?)\n\n', str_operation=str_to_header),
             Quantity(
                 'header',
                 r'(?:GROMACS|Gromacs) (version:[\s\S]+?)\n\n', str_operation=str_to_header),
@@ -996,6 +1014,7 @@ class GromacsParser:
             # last 40% of trajectory
             interval_indices.append(np.arange(n_traj_split)[len(interval_indices[0]) * 3:])
 
+            print('about to calc rdf')
             rdf_results = self.traj_parser.calc_molecular_rdf(n_traj_split=n_traj_split, n_prune=self._frame_rate, interval_indices=interval_indices)
             if rdf_results is not None:
                 sec_rdfs = sec_results.m_create(RadialDistributionFunction)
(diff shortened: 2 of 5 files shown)


atomisticparsers/gromacs/parser.py
```diff
@@ -83,27 +83,46 @@ def str_to_input_parameters(val_in):
...
-            energies = dict()
-            for val in val_in.strip().split('\n'):
-                val = val.strip()
-                if val[0].isalpha():
-                    keys = [k.strip() for k in energy_keys_re.findall(val)]
-                    keys = ['P%s' % k if k.startswith('res') else k for k in keys if k]
+        def str_to_energies(energies_block):
+
+            max_char_name = 14
+            rows = energies_block.strip().split("\n")
```
ladinesa: use splitlines()

atomisticparsers/gromacs/parser.py
```diff
@@ -83,27 +83,46 @@ def str_to_input_parameters(val_in):
...
+            names = [rows[i_row] for i_row in range(len(rows)) if i_row % 2 == 0]
```
ladinesa: maybe you can do this in a single loop

atomisticparsers/gromacs/parser.py
```diff
@@ -83,27 +83,46 @@ def str_to_input_parameters(val_in):
...
+            values = [rows[i_row] for i_row in range(len(rows)) if i_row % 2 != 0]
+
+            names_re = r"(?<!\S)[A-Za-z0-9.()\/-]{1,}(?: [\S][A-Za-z0-9.()\/-]{1,})*(?!\S)"
+            values_re = r"[-+]?\d+\.\d+e[+-]?\d+"
```
ladinesa: please make this a global regex for floats.

atomisticparsers/gromacs/parser.py
```diff
@@ -83,27 +83,46 @@ def str_to_input_parameters(val_in):
...
+            names = [re.findall(names_re, name) for name in names]
```
ladinesa: do this in a single loop

atomisticparsers/gromacs/parser.py
JFRudzinski: A normal for loop and then use append()?

atomisticparsers/gromacs/parser.py
```diff
@@ -83,27 +83,46 @@ def str_to_input_parameters(val_in):
...
-                    for n, key in enumerate(keys):
-
-                        if key == 'Temperature':
-                            energies[key] = float(values[n]) * ureg.kelvin
-                        elif key.startswith('Pres'):
-                            key = key.rstrip(' (bar)')
-                            energies[key] = float(values[n]) * ureg.bar
-                        else:
-                            energies[key] = float(values[n]) / MOL * ureg.kJ
+                    remainder_len = len(name) % max_char_name
```
ladinesa: I am not very happy with this. I think we should simply do 
```python
        def str_to_energies(val_in):
            n_char = len(re.search(r'( +Total Energy)', val_in).group(1))
            energies = {}
            rows = [v for v in val_in.splitlines() if v]
            for n in range(0, len(rows), 2):
                pointer = 0
                while pointer < len(rows[n]):
                    key = rows[n][pointer: pointer + n_chars].strip()
                    value = rows[n + 1][pointer: pointer + n_chars]
                    if key == 'Temperature':
                        energies[key] = float(value) * ureg.kelvin
                    elif key.startswith('Pres'):
                        key = key.rstrip(' (bar)')
                        energies[key] = float(value) * ureg.bar
                    else:
                        energies[key] = float(value) / MOL * ureg.kJ
                    pointer += n_chars
        return energies
```
also change the regex for the energies to 
```python
            Quantity(
                'energies',
                r'Energies \(kJ/mol\).*\n(\s*[\s\S]+?)\n\n',

File: atomisticparsers/gromacs/parser.py [hunk 6]
```diff
@@ -115,11 +134,11 @@ def str_to_step_info(val_in):
         thermo_quantities = [
             Quantity(
                 'energies',
-                r'Energies \(kJ/mol\)\s*([\s\S]+?)\n\n',
+                r'Energies \(kJ/mol\)\s*([\s\S]+?)\nD.* step.* load imb.*|\n\n',
                 str_operation=str_to_energies, convert=False),
             Quantity(
                 'step_info',
-                r'(Step.+\n[\d\.\- ]+)',
+                r'[\r\n]\s*(Step.+\n[\d\.\- ]+)',
```
ladinesa: can we perhaps make [\r\n] as universal regex for line breaks

atomisticparsers/gromacs/parser.py
```diff
@@ -131,9 +150,9 @@ def str_to_step_info(val_in):
             Quantity('execution_path', r'Executable:\s*(.+)'),
             Quantity('working_path', r'Data prefix:\s*(.+)'),
             # TODO cannot understand treatment of the command line in the old parser
-            # Quantity(
-            #     'header',
-            #     r'(?:GROMACS|Gromacs) (2019[\s\S]+?)\n\n', str_operation=str_to_header),
+            Quantity(
+                'header',
+                r'(?:GROMACS|Gromacs) (2019[\s\S]+?)\n\n', str_operation=str_to_header),
```
ladinesa: is this only for v2019?

atomisticparsers/utils/mdanalysis.py
```diff
@@ -397,19 +409,50 @@ def mean_squared_displacement(start: np.ndarray, current: np.ndarray):
...
             return
 
-        atoms_moltypes = self.get('atoms_info', {}).get('moltypes', [])
-        moltypes = np.unique(atoms_moltypes)
-        dt = self.universe.trajectory.dt
         n_frames = self.universe.trajectory.n_frames
-        times = np.arange(n_frames) * dt
-
         if n_frames < 50:
+            self.logger.warning('Less than 50 frames in trajectory, not calculating molecular'
```
ladinesa: Maybe better At least 50 steps required to calculate MSD?

atomisticparsers/utils/mdanalysis.py
```diff
@@ -397,19 +409,50 @@ def mean_squared_displacement(start: np.ndarray, current: np.ndarray):
...
+                                ' mean squared displacements.', UserWarning)
+            return
+
+        dt = getattr(self.universe.trajectory, 'dt')
```
ladinesa: just a new styling thing for cases like this.
```python
if (dt := getattr(self.universe.trajectory, 'dt')) is not None:
    return

File: atomisticparsers/utils/mdanalysis.py [hunk 10]
```diff
@@ -397,19 +409,50 @@ def mean_squared_displacement(start: np.ndarray, current: np.ndarray):
...
+                    moltype_indices = np.array([atom._ix for atom in bead_groups[moltype]._atoms])
+                    molnums = self.universe.atoms.molnums[moltype_indices]
+                    molnum_types = np.unique(molnums)
+                    molnum_types_rnd = np.sort(np.random.choice(molnum_types, size=max_mols))
+                    atom_indices_rnd = moltype_indices[np.concatenate([np.where(molnums == molnum)[0] for molnum in molnum_types_rnd])]
+                    selection = ' '.join([str(i) for i in atom_indices_rnd])
+                    selection = f'index {selection}'
+                    ags_moltype_rnd = self.universe.select_atoms(selection)
+                    bead_groups[moltype] = BeadGroup(ags_moltype_rnd, compound='fragments')
+                    self.logger.warning('The number of molecules of type ' + moltype + ' exceeds the maximum of '
```
ladinesa: We should not put variables in logs. please remove moltype and max_mols . Number of mol types exceed max n of mols for calc... msd.... is sufficient enough.

ladinesa:
I think this is good to merge except for the energies function. Please consider my suggestion, I think it becomes much simpler.

atomisticparsers/gromacs/parser.py
JFRudzinski: To the later point, I just found there was an error in my regex, I just changed it to 
```
            Quantity(
                'energies',
                r'Energies \(kJ/mol\)\s*([\s\S]+?)\n(?:D.* step.* load imb.*|\n\n)',
                str_operation=str_to_energies, convert=False),
```

But I had to add this other term because of some printing "errors" that I found in some of my test log files, e.g.:
```
           Step           Time         Lambda
           4000        8.00000        0.00000

   Energies (kJ/mol)
        LJ (SR)        LJ (LR)   Coulomb (SR)   Coulomb (LR)      Potential
   -1.00701e+03    0.00000e+00    0.00000e+00    0.00000e+00   -1.00701e+03
    Kinetic En.   Total Energy    Temperature Pressure (bar)
    3.96211e+03    2.95511e+03    3.18005e+02    5.51459e+03
DDD  step 4499 load imb.: force  5.1%
```

This can happen where the load imb. line is printed directly under the energy block. Obv. I am open to suggestions if you have a more general regex to avoid grabbing this line.

atomisticparsers/gromacs/parser.py
JFRudzinski: done.

atomisticparsers/gromacs/parser.py
ladinesa: Ah yes, I have completely forgotten about this, please change it to 
```python
Quantity(
                'energies',
                r'Energies \(kJ/mol\).*\n(\s*[\s\S]+?)(?:\n.*step.* load imb.*|\n\n)',

Comment (ladinesa) [reply atomisticparsers/gromacs/parser.py]: This will not cause any problem since it is not captured. We can simply keep on expanding the regex if we find more cases.

File: atomisticparsers/gromacs/parser.py (hunk 2 above)
Comment (ladinesa) [reply atomisticparsers/gromacs/parser.py]: I think this becomes moot.

File: atomisticparsers/gromacs/parser.py (hunk 5 above)
Comment (JFRudzinski) [reply atomisticparsers/gromacs/parser.py]: Actually, my new regex also doesn't work. I find it extremely difficult to implement regex into the parser because the functionality of the regex expression is somehow different than the online tool that I use for testing regex. I think it's something to do with how line breaks are dealt with but I am not sure. If you have any advice about this, it would be much appreciated.

Comment (ladinesa) [reply atomisticparsers/gromacs/parser.py]: I think we change \n to [\n\r] (see my other comment) because windows print \r.

File: atomisticparsers/gromacs/parser.py (hunk 1 above)
Comment (JFRudzinski) [reply atomisticparsers/gromacs/parser.py]: done.

File: atomisticparsers/gromacs/parser.py (hunk 2 above)
Comment (JFRudzinski) [reply atomisticparsers/gromacs/parser.py]: assuming what I said above is correct, then done.

File: atomisticparsers/gromacs/parser.py (hunk 3 above)
Comment (JFRudzinski) [reply atomisticparsers/gromacs/parser.py]: done.

File: atomisticparsers/gromacs/parser.py (hunk 4 above)
Comment (JFRudzinski) [reply atomisticparsers/gromacs/parser.py]: Can you clarify a bit your intent? Do you mean combine the 2 list comprehensions somehow into 1 loop, or also the loop over names below?

Comment (ladinesa) [reply atomisticparsers/gromacs/parser.py]: I think this becomes moot, if you adapt my suggestion below. We should not try to do anymore regex manipulations instead we take advantage of the fixed character length for keys and values.

**Comment by JFRudzinski** [comment]
I am now just thinking about all these values in the Energies block which we are not even storing. Currently we have in `parse_thermodynamic_data():`
```
                # TODO add other energy contributions, properties
                energy_keys = ['LJ (SR)', 'Coulomb (SR)', 'Potential', 'Kinetic En.']

                sec_energy = sec_scc.m_create(Energy)
                for key in thermo_data.keys():
                    val = thermo_data.get(key)[calculation_index]
                    if val is None:
                        continue

                    if key == 'Total Energy':
                        sec_energy.total = EnergyEntry(value=val)
                    elif key == 'Potential':
                        sec_energy.potential = EnergyEntry(value=val)
                    elif key == 'Kinetic En.':
                        sec_energy.kinetic = EnergyEntry(value=val)
                    elif key == 'Coulomb (SR)':
                        sec_energy.coulomb = EnergyEntry(value=val)
                    elif key == 'Pressure':
                        sec_scc.pressure = val
                    elif key == 'Temperature':
                        sec_scc.temperature = val
                    if key in energy_keys:
                        sec_energy.contributions.append(
                            EnergyEntry(kind=self._metainfo_mapping[key], value=val))
```

I can of course extend `energy_keys` but then we will continually have to extend this. Can we not just put all the remaining values in the Energies block? Then if we find other values that don't belong (e.g., a different type of pressure or something) then we can extend the `elif` list instead? This will be simpler I think

**Comment by ladinesa** [comment]
> Can we not just put all the remaining values in the Energies

we can also just have an energies mapping.

**Comment by JFRudzinski** [comment]
> > Can we not just put all the remaining values in the Energies
> 
> we can also just have an energies mapping.

My point is that there can be essentially infinite energy contributions depending on the method. I don't want to keep track of all of these. I just want to throw them into the contributions section, and only keep track of non-energetic contributions that are not being stored. 

Perhaps a map is necessary though. I am now looking at all the quantities which we are not storing. Volume, virial, individual pressure contributions. This is all being read from the edr file but currently not being stored to my knowledge.

**Comment by JFRudzinski** [comment]
> > Can we not just put all the remaining values in the Energies
> 
> we can also just have an energies mapping.

I also don't really understand the point of `contributions` exactly. The separation seems arbitrary: kinetic and potential energy and Coloumb (SR) are not in contributions, but other energy quantities are. Is there some logic here?

**Comment by ladinesa** [comment]
> the

I s

> I just want to throw them into the contributions section, and only keep track of non-energetic contributions that are not being stored.

for me, it is still better to keep track them in a map rather than assuming that all non-registered items are energies esp so that the energies block contain non-energy items.

**Comment by ladinesa** [comment]
> The separation seems arbitrary: kinetic and potential energy and Coloumb (SR) are not in contributions, but other energy quantities are. Is there some logic here

coulomb should be mapped to electrostatic? there are cases where a contribution e.g. electronic can can have kinetic and potential parts. so they are under EnergyEntry, where you can value, potential, kinetic, correction etc

**Comment by JFRudzinski** [comment]
> > the
> 
> I s
> 
> > I just want to throw them into the contributions section, and only keep track of non-energetic contributions that are not being stored.
> 
> for me, it is still better to keep track them in a map rather than assuming that all non-registered items are energies esp so that the energies block contain non-energy items.

Ok, I will try to make a reasonable map I guess.

What do you think about either:
1. only having `total` defined within a subsection (energy or forces), and then all other quantities under that subsection  within contributions OR
2. removing contributions entirely and just listing all quantities under a subsection together?

**Comment by ladinesa** [comment]
> 2\. contributions

contributions are for the ones that we did not define explicitly as quantity.

**Comment by ladinesa** [comment]
> 1. only having `total`

I still think that we should explicitly define energy quantities (total, xc, electronic, electrostatic etc..), and only those that we did not define which we put under contributions.

**Comment by JFRudzinski** [comment]
> e energy quantities (total, xc, electronic, electrostatic etc..), and only those that we did not define which we put under contributions.

Ok, thanks. I will try to make sense of all this and come back with some sort of plan.

**Comment by JFRudzinski** [comment]
@ladinesa I see that you made a note to eventually remove the `thermodynamics` section under `BaseCalculation`. What is your reasoning for this? Where were you planning to put free energies, pressures, volume, etc? directly under `BaseCalculation`?

**Comment by ladinesa** [comment]
> . What is your reasoning for this? Where were you planning to put free energies, pressures, volume, etc? directly under

I remember having this discussion with you, luca and lauri. we decided to simply put them directly under calculation since thermodynamics is not really a single calc property but rather workflow.

**Comment by JFRudzinski** [comment]
> scussion with you, luca and lauri. we decided to simply put them directly under calculation since thermodynamics is not really a sin

ok thanks. I guess I will just populate most quantities directly on calculation for now. I need to discuss with Chema and Nathan about the organization of this class, because right now it is somewhat difficult to navigate, and I believe there are lots of redundant definitions. So, I hope I can work with them to improve this.

I just didn't want to go against your plans unintentionally.

**Comment by JFRudzinski** [comment]
@ladinesa Ok, I finished redoing the parse_thermodynamic_data() function for Gromacs. In summary, I extended the list of keys explicitly dealt with, and added the storage of gromacs specific quantities, for those outside of our list. I also made some changes/additions to the calculation metainfo, which I think coincides for the previous plans to reorganize this section.  

Please let me know what you think.

File: atomisticparsers/gromacs/parser.py [hunk 11]
```diff
@@ -84,28 +86,84 @@ def str_to_input_parameters(val_in):
...
-                        else:
-                            energies[key] = float(values[n]) / MOL * ureg.kJ
+            thermo_common = ['Total Energy', 'Potential', 'Kinetic En.', 'Temperature',
+                             'Pressure (bar)', 'LJ (SR)', 'Coulomb (SR)', 'Bond', 'Proper Dih.']
+            for quant in thermo_common:
+                n_chars_val = re.search(rf'( +{quant})', val_in)
+                n_chars_val = len(n_chars_val.group(1)) if n_chars_val is not None else None
+                if n_chars_val is not None:
+                    break
+            # n_chars_val = len(re.search(r'( +Total Energy)', val_in).group(1))
```
ladinesa: Please remove out commented

atomisticparsers/gromacs/parser.py
```diff
@@ -84,28 +86,84 @@ def str_to_input_parameters(val_in):
...
+                n_chars_val = 15
+            energies = {}
+            rows = [v for v in val_in.splitlines() if v]
+            for n in range(0, len(rows), 2):
+                pointer = 0
+                while pointer < len(rows[n]):
+                    key = rows[n][pointer: pointer + n_chars_val].strip()
+                    value = rows[n + 1][pointer: pointer + n_chars_val]
+                    energies[key] = float(value)
+                    # if key == 'Temperature':
```
ladinesa: here also

atomisticparsers/gromacs/parser.py
```diff
@@ -84,28 +86,84 @@ def str_to_input_parameters(val_in):
...
+                    # elif key.startswith('Pres'):
+                    #     key = key.rstrip(' (bar)')
+                    #     energies[key] = float(value) * ureg.bar
+                    # else:
+                    #     energies[key] = float(value) * ureg.kJ * MOL
+                    pointer += n_chars_val
             return energies
 
+        # def str_to_energies(energies_block):
+
```
ladinesa: here also

atomisticparsers/gromacs/parser.py
```diff
@@ -645,8 +719,9 @@ def parse_thermodynamic_data(self):
             thermo_data = self.energy_parser
 
         calculation_times_ps = thermo_data.get('Time')
-        calculation_times_ps = calculation_times_ps.magnitude * ureg.convert(
-            1.0, calculation_times_ps.units, ureg.picosecond)
+        # calculation_times_ps = calculation_times_ps.magnitude * ureg.convert(
```
ladinesa: not sure what changed here

atomisticparsers/gromacs/parser.py
```diff
@@ -215,15 +273,15 @@ def get_tpstat_settings(self):
...
-        self._energy_keys = [
-            'LJ (SR)', 'Coulomb (SR)', 'Potential', 'Kinetic En.', 'Total Energy',
-            'Vir-XX', 'Vir-XY', 'Vir-XZ', 'Vir-YX', 'Vir-YY', 'Vir-YZ', 'Vir-ZX', 'Vir-ZY',
-            'Vir-ZZ']
-        self._pressure_keys = [
-            'Pressure', 'Pres-XX', 'Pres-XY', 'Pres-XZ', 'Pres-YX', 'Pres-YY', 'Pres-YZ',
-            'Pres-ZX', 'Pres-ZY', 'Pres-ZZ']
-        self._temperature_keys = ['Temperature']
-        self._time_keys = ['Time']
+        # self._energy_keys = [
```
ladinesa: here also

atomisticparsers/gromacs/parser.py
```diff
@@ -245,14 +303,14 @@ def parse(self, key):
...
             val = np.asarray(val)
-        if key in self._energy_keys:
-            val = val / MOL * ureg.kJ
-        elif key in self._temperature_keys:
-            val = val * ureg.kelvin
-        elif key in self._pressure_keys:
-            val = val * ureg.bar
-        elif key in self._time_keys:
-            val = val * ureg.ps
+        # if key in self._energy_keys:
```
ladinesa: here also

atomisticparsers/gromacs/parser.py
```diff
@@ -559,12 +617,28 @@ def __init__(self):
...
         # max cumulative number of atoms for all parsed trajectories to calculate sampling rate
         self._cum_max_atoms = 2500000
+        self._gro_energy_units = ureg.kilojoule * MOL
+        self._thermo_maps = {}
+        self._thermo_maps['ignore_list'] = ['Time', 'Box-X', 'Box-Y', 'Box-Z']
+        self._thermo_maps['dir_to_ind_map'] = {
+            'XX': np.array([0, 0]), 'XY': np.array([0, 1]), 'XZ': np.array([0, 2]),
+            'YX': np.array([1, 0]), 'YY': np.array([1, 1]), 'YZ': np.array([1, 2]),
+            'ZX': np.array([2, 0]), 'ZY': np.array([2, 1]), 'ZZ': np.array([2, 2])}
+        self._thermo_maps['base_calc_map'] = {
```
ladinesa: i do not quite like the idea of putting all the maps in thermo_maps because they have different vallues. Also for the base_calc_map, since this is a fixed key and unit, I suggest using tuple insted of dict.

atomisticparsers/gromacs/parser.py
```diff
@@ -669,30 +744,70 @@ def parse_thermodynamic_data(self):
 
             calculation_index = time_map[time]['calculation_index']
             if calculation_index is not None:
-                # TODO add other energy contributions, properties
-                energy_keys = ['LJ (SR)', 'Coulomb (SR)', 'Potential', 'Kinetic En.']
+                thermo_keys = thermo_data.keys()
+                pressure_tensor = None
+                if any([key.startswith('Pres-') for key in thermo_keys]):
+                    pressure_tensor = np.empty(shape=(3, 3))
```
ladinesa: maybe use np.zeros

atomisticparsers/gromacs/parser.py
```diff
@@ -669,30 +744,70 @@ def parse_thermodynamic_data(self):
...
-                    elif key == 'Pressure':
-                        sec_scc.pressure = val
-                    elif key == 'Temperature':
-                        sec_scc.temperature = val
-                    if key in energy_keys:
-                        sec_energy.contributions.append(
-                            EnergyEntry(kind=self._metainfo_mapping[key], value=val))
+                    # Attributes of BaseCalculation
+                    if key in thermo_maps['base_calc_map'].keys():
+                        setattr(sec_scc, thermo_maps['base_calc_map'][key]['nomad_key'], val * thermo_maps['base_calc_map'][key]['gro_units'])
```
ladinesa: use set_scc.m_set(sec_scc.m_get_quantity_definition(thermo_maps['base_calc_map'][key]['nomad_key]), val * thermo_maps['base_calc_map'][key]['gro_units'])

atomisticparsers/gromacs/parser.py
```diff
@@ -669,30 +744,70 @@ def parse_thermodynamic_data(self):
...
+                    # accumulating tensor quantites
+                    elif key.startswith('Pres-'):
+                        dir = key.split('-')[1]
+                        print(dir)
```
ladinesa: remove prints

atomisticparsers/gromacs/parser.py
```diff
@@ -559,12 +617,28 @@ def __init__(self):
...
-        self._metainfo_mapping = {
-            'LJ (SR)': 'Leonard-Jones', 'Coulomb (SR)': 'coulomb',
-            'Potential': 'potential', 'Kinetic En.': 'kinetic'}
         self._frame_rate = None
         # max cumulative number of atoms for all parsed trajectories to calculate sampling rate
         self._cum_max_atoms = 2500000
+        self._gro_energy_units = ureg.kilojoule * MOL
+        self._thermo_maps = {}
+        self._thermo_maps['ignore_list'] = ['Time', 'Box-X', 'Box-Y', 'Box-Z']
+        self._thermo_maps['dir_to_ind_map'] = {
```
ladinesa: maybe do tensor_keys = np.array([['XX, 'XY', 'XZ'], [YX], ...]) and do tensor[np.where(tensor_keys=key)] = val

atomisticparsers/gromacs/parser.py
```diff
@@ -669,30 +744,70 @@ def parse_thermodynamic_data(self):
...
+                        if ind is not None:
+                            pressure_tensor[ind] = val * ureg.bar
+                    elif key.startswith('Vir-'):
+                        dir = key.split('-')[1]
+                        ind = thermo_maps['dir_to_ind_map'].get(dir)
+                        if ind is not None:
+                            virial_tensor[ind] = val * (ureg.bar * ureg.nm**3)
+                    # well-defined, single Energy quantities
+                    elif key in thermo_maps['energy_map'].keys():
+                        setattr(sec_energy, thermo_maps['energy_map'][key], EnergyEntry(value=val * self._gro_energy_units))
```
ladinesa: do sec_energy.m_add_sub_section(getattr(Energy, energy_map[key]), EnergyEntry(value=val * unit))

atomisticparsers/gromacs/parser.py
```diff
@@ -669,30 +744,70 @@ def parse_thermodynamic_data(self):
...
+                    # well-defined, piecewise energy quantities
+                    elif key in thermo_maps['vdw_map'].keys():
```
ladinesa: you do not have to do .keys()

atomisticparsers/gromacs/parser.py
```diff
@@ -669,30 +744,70 @@ def parse_thermodynamic_data(self):
...
+                        vdw_dict[thermo_maps['vdw_map'][key]] = val * self._gro_energy_units
+                    elif key in thermo_maps['electrostatic_map'].keys():
+                        electrostatic_dict[thermo_maps['electrostatic_map'][key]] = val * self._gro_energy_units
+                    else:
+                        # try to identify other known energy keys to be stored as gromacs-specific
+                        if any([keyword in key.lower() for keyword in thermo_maps['energy_contains_list']]):
+                            sec_energy.contributions.append(
+                                EnergyEntry(kind='x_gromacs_' + key, value=val * self._gro_energy_units))
+                        else:  # store all other quantities as gromacs-specific under BaseCalculation
+                            setattr(sec_scc, 'x_gromacs_' + key, val)  # TODO Need to create metainfo for these or store using Chema's method
```
ladinesa: else if hasattr(sec_scc, 'x_gromacs_{key}'):
    sec_scc.m_set(sec_scc.m_get_quantity_definition(x_gromacs_{key}), val)

atomisticparsers/gromacs/parser.py
```diff
@@ -669,30 +744,70 @@ def parse_thermodynamic_data(self):
...
+
+                sec_scc.pressure_tensor = pressure_tensor if pressure_tensor is not None else None
```
ladinesa: you can get rid of the if

atomisticparsers/gromacs/parser.py
```diff
@@ -669,30 +744,70 @@ def parse_thermodynamic_data(self):
...
+                sec_scc.virial_tensor = virial_tensor if virial_tensor is not None else None
+                if vdw_dict:
+                    total = sum(val for _, val in vdw_dict.items())
+                    sec_energy.van_der_waals = EnergyEntry(value=total, short_range=vdw_dict.get('short_range'),
```
ladinesa, line 722: don't we have units for the values?

atomisticparsers/utils/mdanalysis.py
```diff
@@ -220,10 +224,16 @@ def get_rdf_avg(rdf_results_tmp, rdf_results, interval_indices, n_frames_split):
...
-        if bead_groups is None:
-            return {}
+        if bead_groups is {}:
+            return bead_groups
+        moltypes = [moltype for moltype in bead_groups.keys()]
+        del_list = []
+        for i_moltype, moltype in enumerate(moltypes):
+            if bead_groups[moltype]._nbeads > max_mols:
+                del_list.append(i_moltype)
+                self.logger.warning('The number of molecules of type ' + moltype + ' exceeds the maximum of '
```
ladinesa: please remove variables in warning

atomisticparsers/utils/mdanalysis.py
```diff
@@ -397,19 +409,48 @@ def mean_squared_displacement(start: np.ndarray, current: np.ndarray):
...
+                    molnum_types_rnd = np.sort(np.random.choice(molnum_types, size=max_mols))
+                    atom_indices_rnd = moltype_indices[np.concatenate([np.where(molnums == molnum)[0] for molnum in molnum_types_rnd])]
+                    selection = ' '.join([str(i) for i in atom_indices_rnd])
+                    selection = f'index {selection}'
+                    ags_moltype_rnd = self.universe.select_atoms(selection)
+                    bead_groups[moltype] = BeadGroup(ags_moltype_rnd, compound='fragments')
+                    self.logger.warning('Maximum number of molecules for calculating the msd has been reached.'
+                                        ' Will make a random selection for calculation.')
+                except Exception:
+                    self.logger.warning('Tried to select random molecules for large group ' + moltype
```
ladinesa: please remove variable in warning

tests/data/gromacs/cgwater/mdrun.log
```diff
@@ -2547,7 +2547,6 @@ Writing checkpoint, step 100000 at Mon Apr  4 10:52:48 2022
 	Statistics over 100001 steps using 1001 frames
 
    Energies (kJ/mol)
-        LJ (S   Energies (kJ/mol)
```
ladinesa, line 2550: please add test for added energy types

tests/test_gromacsparser.py
```diff
@@ -209,7 +213,7 @@ def test_geometry_optimization(parser):
     assert sec_workflow.results.optimization_steps == 12
     assert sec_workflow.method.optimization_steps_maximum == 5000
     assert len(sec_workflow.results.energies) == 11
-    assert sec_workflow.results.energies[2].magnitude == approx(8.244726173423345e-17)
+    assert sec_workflow.results.energies[2].magnitude == approx(2.9900472759121395e+31)
```
ladinesa, line 232: this seems to be not correct? units?

atomisticparsers/gromacs/parser.py
```diff
@@ -93,7 +93,6 @@ def str_to_energies(val_in):
                 n_chars_val = len(n_chars_val.group(1)) if n_chars_val is not None else None
```
ladinesa: we can get rid of the for loop if we do re.search(rf'( +{"|".join(thermo_common)})', ' val_in)

atomisticparsers/gromacs/parser.py
```diff
@@ -621,24 +546,24 @@ def __init__(self):
         # max cumulative number of atoms for all parsed trajectories to calculate sampling rate
         self._cum_max_atoms = 2500000
         self._gro_energy_units = ureg.kilojoule * MOL
-        self._thermo_maps = {}
-        self._thermo_maps['ignore_list'] = ['Time', 'Box-X', 'Box-Y', 'Box-Z']
-        self._thermo_maps['dir_to_ind_map'] = {
+        self._thermo_ignore_list = ['Time', 'Box-X', 'Box-Y', 'Box-Z']
+        # self._tensor_keys = np.array([['XX', 'XY', 'XZ'], ['YX', 'YY', 'YZ'], ['ZX', 'ZY', 'ZZ']])
+        self._dir_to_ind_map = {
```
ladinesa: please rename to self.tensor_index_map

atomisticparsers/gromacs/parser.py
```diff
@@ -621,24 +546,24 @@ def __init__(self):
...
+            'Volume': ('volume', ureg.nm**3),
+            'Density': ('density', ureg.kilogram / ureg.m**3),
+            'Pressure (bar)': ('pressure', ureg.bar),
+            'Pressure': ('pressure', ureg.bar),
+            'Enthalpy': ('enthalpy', self._gro_energy_units)
+        }
+        self._energy_map = {'Potential': 'potential', 'Kinetic En.': 'kinetic', 'Total Energy': 'total', 'pV': 'pressure_volume_work'}
+        self._vdw_map = {'LJ (SR)': 'short_range', 'LJ (LR)': 'long_range', 'Disper. corr.': 'correction'}
+        self._electrostatic_map = {'Coulomb (SR)': 'short_range', 'Coul. recip.': 'long_range'}
+        self._energy_contains_list = ['bond', 'angle', 'dih.', 'coul-', 'coulomb-', 'lj-', 'en.']
```
ladinesa: rename this perhaps to _energy_keys?

JFRudzinski:
@ladinesa I tried to resolve all conversations that were already addressed. The remaining items are things I am asking for your confirmation on.

A couple questions:
1. I always forget, do we merge this first or the NOMAD branch first? (Also, can you check my corresponding MR on gitlab?)
2. These corrections were all based on a users post. I would like to update them and give them an ETA on when these changes will be available in the beta. How long does this take usually from the time we merge both branches?

ladinesa:
Can you mark the each review point as resolved if they are? or comment?

1. We merge this first then we update nomad.
2. Usually it takes about a couple of weeks before the changes appear on production-dev.

atomisticparsers/gromacs/metainfo/gromacs.py
```diff
@@ -2019,10 +2019,39 @@ class Constraint(simulation.system.Constraint):
         repeats=True,)
 
 
+class Entry(MSection):
```
ladinesa, line 2022: I think EnergyEntry already has these defintions.

atomisticparsers/gromacs/parser.py
```diff
@@ -84,26 +86,21 @@ def str_to_input_parameters(val_in):
...
-                        if key == 'Temperature':
-                            energies[key] = float(values[n]) * ureg.kelvin
-                        elif key.startswith('Pres'):
-                            key = key.rstrip(' (bar)')
-                            energies[key] = float(values[n]) * ureg.bar
-                        else:
-                            energies[key] = float(values[n]) / MOL * ureg.kJ
+            thermo_common = [r'Total Energy', r'Potential', r'Kinetic En.', r'Temperature',
+                             r'Pressure \(bar\)', r'LJ \(SR\)', r'Coulomb \(SR\)', r'Proper Dih.']
+            n_chars_val = re.search(rf'( +{"| +".join(thermo_common)})', val_in)
```
ladinesa, line 91: I think the re pattern is wrong, we do not want to eat into the spaces after the key. should only be "|".join(thermo_common)

atomisticparsers/gromacs/parser.py
```diff
@@ -669,30 +662,70 @@ def parse_thermodynamic_data(self):
...
+                    elif key in self._vdw_map:
+                        vdw_dict[self._vdw_map[key]] = val * self._gro_energy_units
+                    elif key in self._electrostatic_map:
+                        electrostatic_dict[self._electrostatic_map[key]] = val * self._gro_energy_units
+                    else:
+                        # try to identify other known energy keys to be stored as gromacs-specific
+                        if any([keyword in key.lower() for keyword in self._energy_keys_contain]):
+                            sec_energy.contributions.append(
+                                EnergyEntry(kind='x_gromacs_' + key, value=val * self._gro_energy_units))
+                        else:  # store all other quantities as gromacs-specific under BaseCalculation
```
ladinesa, line 714: please see my suggestion for this.

atomisticparsers/gromacs/parser.py
```diff
@@ -1075,6 +1108,7 @@ def reuse_parser(self, parser):
         self.log_parser.quantities = parser.log_parser.quantities
 
     def parse(self, filepath, archive, logger):
+
```
ladinesa: remove line break

JFRudzinski:
@ladinesa I am considering now storing the gromacs-specific energies and thermo data in a json format rather than a list. I included both options in my last push. Could you let me know if there is a preference here?

ladinesa:
> @ladinesa I am considering now storing the gromacs-specific energies and thermo data in a json format rather than a list. I included both options in my last push. Could you let me know if there is a preference here?

I would actually like to explicitly define them, maybe we put parameters as json but not energies?

JFRudzinski:
> > @ladinesa I am considering now storing the gromacs-specific energies and thermo data in a json format rather than a list. I included both options in my last push. Could you let me know if there is a preference here?
> 
> I would actually like to explicitly define them, maybe we put parameters as json but not energies?

In what sense do you mean "explicitly define them"? 

If the way that I have implemented them into contributions is what you mean, then I understand. The benefit of that is that you always know there is a kind and a value. I was considering storing in json because I am working on more custom uploading of MD data, and I thought it might be nicer in this way for consistency. In the custom approach, you can upload calculations with arbitrary attributes, which will then be most easily stored within the json format. However, I am not set on this for the gromacs quantities, it was just a thought.

However, if you mean that we make a list of all possible keys and make metainfo `x_gromacs_key`, this is impossible, as the key names can depend on user naming within the force field.

ladinesa:
> However, if you mean that we make a list of all possible keys and make metainfo `x_gromacs_key`, this is impossible, as the key names can depend on user naming within the force field.

yes this was what i was thinking, but if you think adding them as contributions is better, it is also equally ok. Can you please link me to the custom MD data you are working on so I can understand more what you are trying to accomplish?

JFRudzinski:
> s thinking, but if you think adding them as contributions is better, it is also equally ok. Can you please link me to the custom MD data you are working on so I can understand more what you are trying to accomplish?

Ok, I will keep them as contributions then. I think it is really the only way in this case for many of the quantities, and I think spending a bunch of time on this type of thing is not very useful since they will very rarely be used. I think if a quantity is being used often then we add it to the main metainfo.

I will write you about the custom MD stuff on the corresponding issue in a minute.

PR #59, opened 2023-08-09
32 md parsers superclass

### LINKED ISSUE #32: Super class for MD parsers
We can perhaps make a super class for MD parsers, which deals with repeated functions like parts of parse_thermodynamic_data() and the frame rate sampling.

ladinesa:
Closes #32


ladinesa:
I implemented an initial version of the md parser class and applied it to lammps. @JFRudzinski please have a look at it and tell me what you think. thanks.

atomisticparsers/lammps/parser.py
```diff
@@ -663,44 +648,35 @@ def parse_thermodynamic_data(self):
...
-                    elif key == 'toteng':
-                        sec_energy.current = EnergyEntry(value=val[calculation_index])
-                        sec_energy.total = EnergyEntry(value=val[calculation_index])
-                    elif key == 'press':
-                        sec_scc.pressure = val[calculation_index]
-                    elif key == 'temp':
-                        sec_scc.temperature = val[calculation_index]
-                    elif key == 'cpu':
-                        sec_scc.time_calculation = float(val[calculation_index])
+        for step in self.thermodynamics_steps:
```
JFRudzinski, line 652: I think you are assuming here that trajectory_steps is a subset of thermodynamics_steps. You have to somehow take the union of these lists to create the calculation entries.

atomisticparsers/lammps/parser.py
```diff
@@ -965,14 +941,14 @@ def parse_workflow(self):
             interval_indices.append(np.arange(n_traj_split)[len(interval_indices[0]) * 3:])
 
             # calculate molecular radial distribution functions
```
JFRudzinski: Somewhat unrelated, but it is relevant to simplifying the parser code:

We still have 2 separate implementations of calculating the rdf and msd. One within the parser and one within the workflow normalizer. I know the calculation is not being performed twice, but it would still be nice to remove this from the parser when we can. Is it already safe to do this? I feel like the workflow2 section is relatively active now, but you know better?

atomisticparsers/lammps/parser.py
JFRudzinski: ok, I see now that in the super parser the steps function is the union of the lists. But it is still unclear to me that this corresponds to self.thermodynamics_steps, since on line 1330 we have:
`        self.thermodynamics_steps = [int(n) for n in thermo_data.get('Step', [])]`

Please just check this and confirm.

atomisticparsers/utils/parsers.py
```diff
@@ -0,0 +1,169 @@
...
+        return np.amax(self._info.get('n_atoms', [0]))
+
+    @n_atoms.setter
+    def n_atoms(self, value):
+        self._info['n_atoms'] = [value] if not isinstance(value, Iterable) else value
+
+    @property
+    def frame_rate(self) -> int:
+        '''
+        Returns the sampling rate of saved thermodynamics data and trajectory.
```
JFRudzinski, line 134: Why do you say both thermodynamics and trajectory. We should only be pruning the trajectory (i.e., system) if the data is too heavy, right?

Also, maybe the comment is misleading cause it sounds like you are returning the sampling rate from the simulation, not the sampling rate for the archive. maybe clarify?

JFRudzinski (approved):
Looks great @ladinesa , I only have minor comments. I think this is a really nice step to simplifying these parsers. Thanks for your efforts!

atomisticparsers/lammps/parser.py
ladinesa: the setter function takes in the full thermo steps while the getter funtion returns the sampled steps. same thing for the trajectory. steps correspond to the sampled set of the union of thermo and traj

atomisticparsers/lammps/parser.py
ladinesa: You mean the msd calc is done in parser and will only calc in normalizer if not yet calculated? On a relevant note, I did not touch the gromacs parser yet since I am waiting for your other mr.

atomisticparsers/utils/parsers.py
ladinesa: > Also, maybe the comment is misleading cause it sounds like you are returning the sampling rate from the simulation, not the sampling rate for the archive. maybe clarify?

ah yes you are completely right here, indeed this is the sampling rate for the archive.

ladinesa: > Why do you say both thermodynamics and trajectory. We should only be pruning the trajectory (i.e., system) if the data is too heavy, right

frame rate is calculated from _steps (union of thermo and traj) .  Should we also not sample thermo, maybe in the future there will be large calc data? Regarding large data, I plan to dump large arrays to hdf5. Do you also want to dump the full trajectory to hdf5?

atomisticparsers/lammps/parser.py
ladinesa: I will be removing msd and rdf calc in lammps, do i simply remove the tests? or do I try to transfer it to workflow norm test?

ladinesa: Ah so they are alreay there. then I will simply remove the calculation part.

atomisticparsers/lammps/parser.py
JFRudzinski: ok, great :+1:

atomisticparsers/lammps/parser.py
JFRudzinski: Yeah, so just to clarify (I think you see now): The workflow normalizer (workflow2) does the same calculation in case these sections are not already defined. We did that previously while workflow2 was still under initial development. But once workflow2 is fully functional (which I think is already the case). We can completely remove these calculations from the parsers (Lammps and Gromacs).

atomisticparsers/utils/parsers.py
JFRudzinski: For the frame_rate: 

Oh, ok I see now. Yeah, I don't think this makes sense in the current implementation. The frame_rate should only be calculated from the number of trajectory frames, since the criterion is based off the number of cum nr. of atoms in the trajectory.

It is conceivable that in the future there could be some higher dim quantity that is stored in calc and thus limits the frame_rate of that section, but that will have to be a separate criterion anyway, so we should just wait and address that as it comes I think.

JFRudzinski: For the hdf5 storage: By "dump the full trajectory" do you mean the positions or the full system list of the archive or something else?

ladinesa: all the positions that were parsed.

ladinesa: > The frame_rate should only be calculated from the number of trajectory frames, since the criterion is based off the number of cum nr. of atoms in the trajectory.

i do not follow this, then we do not need steps and also we do not need to sample thermo_steps?

JFRudzinski: > do not need ste

I don't really think this is necessary at the moment. There are several 3rd party modules like MDAnalysis that can read a host of output files from simulations to get the positions. If we dump to hdf5 we are just duplicating the position data that is held in the raw files (again, which are typically easily read) plus we also then have a subset again in the archive. 

Maybe in the future if someone finds that this would be easier. But it's hard for me to imagine that anyone would actually use this rather than going to the raw files. What do you think?

ladinesa: This was exactly what I thought before also but if we have save them in a standard format then we can leverage also the h5web visualizer in nomad.

JFRudzinski: > > The frame_rate should only be calculated from the number of trajectory frames, since the criterion is based off the number of cum nr. of atoms in the trajectory.
> 
> i do not follow this, then we do not need steps and also we do not need to sample thermo_steps?

I am not 100% sure that I follow the code correctly, so I don't want to tell you which variables you need or don't need.

But here is how I see the overall process:
Data in raw files:
    system_data stored every nth simulation step -- positions, atom_attributes, etc.
    calculation_data stored every mth simulation step -- thermo data, energies, etc.

Parsing strategy:
    1.  for large systems / long trajectories, prune the system_data (only) via frame_rate, which is calculated from cum_atoms = n_frames * n_atoms. n_frames here refers to the total number of steps stored (in the raw data) for system_data only.
    2. create calculation entries for all steps stored (union of stored steps for system_data and calculation_data). note - these step lists may be non-overlapping even without any pruning (nr. 1) due to the storage frequency above.

Does this help? Or which aspect is still unclear?

ladinesa: 1. it is cleaner now, we calculate sampling_rate only from n_traj
2. we do not sample thermo, but i do not get why we need a union of thermo + system steps.

JFRudzinski: > This was exactly what I thought before also but if we have save them in a standard format then we can leverage also the h5web visualizer in nomad.

I am definitely not against it if you think it is useful. Still not sure if the visualizer is useful if you only have positions in there, but maybe I don't understand the functionality.

Would you then not store the positions directly in the archive and just reference the hdf5? I imagine that the referencing could be the beneficial aspect here.

ladinesa: I also have not used it but it seems you do some array manipulations. it is not efficient to store large arrays in the archive directly. We can deal with storing the full trajectory later when the need arises.

JFRudzinski: that sounds like a reasonable plan to me

JFRudzinski:
@ladinesa changes to the gromacs parser look good to me. I don't see anything problematic at the moment. 

If you would like, once you are done (not sure if that is now), I can run some more extended tests to double check everything?

ladinesa:
> @ladinesa changes to the gromacs parser look good to me. I don't see anything problematic at the moment.
> 
> If you would like, once you are done (not sure if that is now), I can run some more extended tests to double check everything?

Thanks for looking at it. I will simply do the same for the other parsers.

ladinesa:
> 



> @ladinesa changes to the gromacs parser look good to me. I don't see anything problematic at the moment.
> 
> If you would like, once you are done (not sure if that is now), I can run some more extended tests to double check everything?

It is pretty much done now for all the atomistic parsers. I did only simple testing of the traj sampling. It would be good if you can verify it it really works as intended especially the referencing of the system from calculation.

@JFRudzinski  @JosePizarro3  please also give me a feedback on the parse_section functionality of the AtomisticParser. It essentially makes parsing simpler as you only needs to provide now the data to be written to the metainfo section. This means we do not anymore need to import the metainfo definitions explicitly. I am not totally sure about the performance as this now involves on-the-fly resolution of the section and quantity definitions.  I did basic testing and it seems to be 5% slower than when we do direct assignment. Anyway if you are happy with this, I will  put this in nomad.parsing. My plan is to eventually have all parsers use this so we do not need to always have a problem that the definitions are not up to date everytime we change.

JosePizarro3:
> > 
> 
> > @ladinesa changes to the gromacs parser look good to me. I don't see anything problematic at the moment.
> > If you would like, once you are done (not sure if that is now), I can run some more extended tests to double check everything?
> 
> It is pretty much done now for all the atomistic parsers. I did only simple testing of the traj sampling. It would be good if you can verify it it really works as intended especially the referencing of the system from calculation.
> 
> @JFRudzinski @JosePizarro3 please also give me a feedback on the parse_section functionality of the AtomisticParser. It essentially makes parsing simpler as you only needs to provide now the data to be written to the metainfo section. This means we do not anymore need to import the metainfo definitions explicitly. I am not totally sure about the performance as this now involves on-the-fly resolution of the section and quantity definitions. I did basic testing and it seems to be 5% slower than when we do direct assignment. Anyway if you are happy with this, I will put this in nomad.parsing. My plan is to eventually have all parsers use this so we do not need to always have a problem that the definitions are not up to date everytime we change.

This looks promising. Nevertheless, this week I am a pretty busy, but I would like to take a look into depth for this utility. Do you mind waiting for next week for my comments?

Thanks!

JFRudzinski:
> > 
> 
> > @ladinesa changes to the gromacs parser look good to me. I don't see anything problematic at the moment.
> > If you would like, once you are done (not sure if that is now), I can run some more extended tests to double check everything?
> 
> It is pretty much done now for all the atomistic parsers. I did only simple testing of the traj sampling. It would be good if you can verify it it really works as intended especially the referencing of the system from calculation.
> 
> @JFRudzinski @JosePizarro3 please also give me a feedback on the parse_section functionality of the AtomisticParser. It essentially makes parsing simpler as you only needs to provide now the data to be written to the metainfo section. This means we do not anymore need to import the metainfo definitions explicitly. I am not totally sure about the performance as this now involves on-the-fly resolution of the section and quantity definitions. I did basic testing and it seems to be 5% slower than when we do direct assignment. Anyway if you are happy with this, I will put this in nomad.parsing. My plan is to eventually have all parsers use this so we do not need to always have a problem that the definitions are not up to date everytime we change.

@ladinesa The `parse_section` functionality appears good to me. I will do as many tests as I can today and tomorrow, since I am on vacation next week.

atomisticparsers/gromacs/parser.py
```diff
@@ -937,139 +893,79 @@ def parse_workflow(self):
...
             value = thermostat_map.get(thermostat, [val for key, val in thermostat_map.items() if key in thermostat])
             value = value if not isinstance(value, list) else value[0] if len(value) != 0 else None
-            workflow.method.thermostat_parameters.thermostat_type = value
+            thermostat_parameters = {}
+            thermostat_parameters['thermostat_type'] = value
             if 'sd' in integrator:
-                workflow.method.thermostat_parameters.thermostat_type = 'langevin_goga'
-            if workflow.method.thermostat_parameters.thermostat_type:
-                flag_thermostat = True
+                thermostat_parameters.thermostat_type = 'langevin_goga'
```
JFRudzinski: This should be `thermostat_parameters['thermostat_type']`

JFRudzinski:
@ladinesa FYI - the majority of my tests worked fine. There are some remaining issues with the Lammps parser, but I would not worry about these, I don't think they are necessary due to your changes. I need to sit down in within the next month or so and try to re-vamp this parser a bit.

ladinesa:
should i merge this then?

JFRudzinski:
> should i merge this then?

Yeah I think so

PR #62, opened 2023-08-23
Add calculation time

ladinesa:
Parse physical and calculation time for calculation frames for parsers that provide timing info.


ladinesa:
@JFRudzinski can you please have a look at this?

PR #67, opened 2023-09-20
Fix system_ref in gromacs

ladinesa:


ladinesa:
@JFRudzinski are you planning to merge the update to the thermostart parameters metainfo soon? otherwise, we should defer the lastest merge on fixing the lammps test.

JFRudzinski:
@ladinesa I would like to merge it ASAP, but I thought we had to fix the pytest first? If that is not the case, please let me know and I will go ahead and merge.

ladinesa:
> @ladinesa I would like to merge it ASAP, but I thought we had to fix the pytest first? If that is not the case, please let me know and I will go ahead and merge.

which test is still failing? the one un atomistic parsers should be fixed once you update nomad. have you fixed test_workflow test in nomad ?

JFRudzinski:
I think I got confused because of this issue with .log file not being there. The test_workflow is still broken. I have to look again to see what the problem is (Sorry this week is very busy for me as I am organizing a CECAM workshop that takes place next week)

JFRudzinski:
Oh no, sorry, I remember now. the test_workflow is not working because of the system_ref problem.

So, should I merge this and then we merge the atomistic branch and then I will rerun the pipeline?

ladinesa:
I think the problem with test workflow is that the code that was removed from the lammps parser that calculates msd is a bit different from the one in normalizing. I have seen you fix this i think. We should merge all our fixes in atomistic. can you please have a look at my fix for the system_ref issue so i can merge it also? then we try to see if these fixes work

JFRudzinski:
> my fix for the system_ref issue so i can merge it also? then we try to see if these

No, the test_workflow error that I am looking at is about the rg calculation. It only is populated for the first entry in the calculation list, because it relies on the system_ref to assign the rg results. This should get fixed by your corrections.

Yes, I will look at your code now, sorry.

Do we merge NOMAD first?

JFRudzinski:
Ok, yes, I think your corrections should do the trick, let's try it!

JFRudzinski:
@ladinesa I am going to merge the NOMAD changes, yeah?

ladinesa:
> @ladinesa I am going to merge the NOMAD changes, yeah?

i think we merge first the atomistic parsers because nomad will not work if the parser is not updated.

JFRudzinski:
but the parser will also not work if NOMAD is not updated, because of the metainfo...but maybe that is less important

ladinesa:
> but the parser will also not work if NOMAD is not updated, because of the metainfo...but maybe that is less important

we will simply rerun the tests again once nomad develop is updated.

PR #73, opened 2023-10-31
72 add bond list and generalize interactions in gromacs and lammps

### LINKED ISSUE #72: Add bond list and generalize interactions in gromacs and lammps
- bond list will be added to system[0] for visualization purposed
 - move to a hierarchical grouping of interactions by type

JFRudzinski:
- Interactions are now grouped into a hierarchy by type, this makes it more straightforward to grab the bond list, and is more intuitive storage for users
- Global bond list is stored under atoms (in system[0] for now).
- Bond list metadata is also available in atoms_group, although not utilized yet

### DIFF
--- a/atomisticparsers/utils/parsers.py
+++ b/atomisticparsers/utils/parsers.py
@@ -202,12 +202,12 @@ class MDParser(AtomisticParser):
         self.parse_section(data, sec_workflow)
         self.archive.workflow2 = sec_workflow
 
-    def parse_interactions(self, interactions: Dict[str, Any], sec_model: MSection) -> None:
+    def parse_interactions(self, interactions: List[Dict], sec_model: MSection) -> None:
 
-        interaction_key_list = Interaction.__dict__.keys()
+        interaction_key_list = [
+            'type', 'name', 'n_inter', 'n_atoms', 'atom_labels', 'atom_indices',
+            'functional_form', 'n_parameters', 'parameters']
         interaction_dict = {}
-        interaction_keys_remove = ['__module__', '__doc__', 'm_def']
-        interaction_key_list = [key for key in interaction_key_list if key not in interaction_keys_remove]
         for interaction_key in interaction_key_list:
             interaction_dict[interaction_key] = np.array([interaction.get(interaction_key) for interaction in interactions], dtype=object)
         interaction_dict = {key: val for key, val in interaction_dict.items()}
@@ -230,12 +230,8 @@ class MDParser(AtomisticParser):
                     interaction_vals = interaction_vals.tolist()
                 elif key == 'n_atoms':
                     interaction_vals = interaction_vals[0]
-                try:
-                    setattr(sec_interaction, key, interaction_vals)
-                except Exception:
-                    self.logger.warning(
-                        'Some issue trying to store ' + key + 'in Interactions section.'
-                        ' Possibly a data type problem. Ignoring these values.')
+                if hasattr(sec_interaction, key):
+                    sec_interaction.m_set(sec_interaction.m_get_quantity_definition(key), interaction_vals)
 
-            if not sec_interaction.get('n_atoms'):
+            if not sec_interaction.n_atoms:
                 sec_interaction.n_atoms = len(sec_interaction.get('atom_indices')[0]) if sec_interaction.get('atom_indices') is not None else None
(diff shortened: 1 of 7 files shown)


atomisticparsers/utils/parsers.py
```diff
@@ -200,3 +201,41 @@ def parse_md_workflow(self, data: Dict[str, Any]) -> None:
         sec_workflow = MolecularDynamics()
         self.parse_section(data, sec_workflow)
         self.archive.workflow2 = sec_workflow
+
+    def parse_interactions(self, interactions: Dict[str, Any], sec_model: MSection) -> None:
+
+        interaction_key_list = Interaction.__dict__.keys()
+        interaction_dict = {}
+        interaction_keys_remove = ['__module__', '__doc__', 'm_def']
+        interaction_key_list = [key for key in interaction_key_list if key not in interaction_keys_remove]
```
ladinesa: you can use the inspect module here, inspect.getmembers(Interaction). But I suggest you simply list down all the class attributes since from line 222-232, you also have key-specific actions.

atomisticparsers/utils/parsers.py
```diff
@@ -200,3 +201,41 @@ def parse_md_workflow(self, data: Dict[str, Any]) -> None:
...
+                if type(interaction_vals[0]).__name__ == 'ndarray':
+                    interaction_vals = np.array([vals.tolist() for vals in interaction_vals], dtype=object)
+                if interaction_vals.all() is None:
+                    continue
+                if key == 'parameters':
+                    interaction_vals = interaction_vals.tolist()
+                elif key == 'n_atoms':
+                    interaction_vals = interaction_vals[0]
+                try:
+                    setattr(sec_interaction, key, interaction_vals)
```
ladinesa: I would do
```python
if hasattr(sec_interaction, key):
   sec_interaction.m_set(sec_interaction.m_get_quantity_defintion(key), val)

File: atomisticparsers/utils/parsers.py [hunk 3]
```diff
@@ -200,3 +201,41 @@ def parse_md_workflow(self, data: Dict[str, Any]) -> None:
...
+                except Exception:
+                    self.logger.warning(
```
ladinesa: I do not like the try here, the only reason that it would fail if there is a data type mismatch, but I would make sure that I pass the right data type in the parser itself. In case, you want to keep this, remove 'key' in log. logger warnings should not be unique otherwise it would clog the logger.

atomisticparsers/utils/parsers.py
```diff
@@ -200,3 +201,41 @@ def parse_md_workflow(self, data: Dict[str, Any]) -> None:
...
+                        'Some issue trying to store ' + key + 'in Interactions section.'
+                        ' Possibly a data type problem. Ignoring these values.')
+
+            if not sec_interaction.get('n_atoms'):
```
ladinesa: I am not sure if there is an overhead with the get method but I would directly get attr, e.g sec_interaction.n_atoms. same with atom_indices

ladinesa:
Enerything looks good, just some tiny comments for your consideration.

atomisticparsers/utils/parsers.py
JFRudzinski: done.

atomisticparsers/utils/parsers.py
JFRudzinski: done.

atomisticparsers/utils/parsers.py
JFRudzinski: removed the try. done.

atomisticparsers/utils/parsers.py
JFRudzinski: I thought it would be nicer to get these automatically. The key-specific actions are only a small subset which have special data structures. But I don't see any easy way to only grab the keys that I want with inspect, and I also agree that this current way is not very nice. So, I just made a hard-coded list as suggested. If you have any other ideas let me know, otherwise I will leave it like this.

JFRudzinski:
@ladinesa Can you also have a look at the corresponding [changes in NOMAD](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/1526) so I can merge that first.

atomisticparsers/utils/parsers.py
ladinesa: normally you do not access __dict__.  you can do this so you do not have to remove keys.
```python
interaction_keys = [n for n, q in inspect.getmembers(Interaction) if isinstance(q, Quantity)]

ladinesa: or if you do not need the keys, simply do
```python
if hasattr(sec_interaction, key):
    sec_interaction.m_set(sec_interaction.m_get_quantity_definition(key), val)

JFRudzinski: Ok, thank you. I have went ahead and used your suggestion with `inspect` now...done.

PR #78, opened 2023-11-28
Remove workflow schema from nomad

ladinesa:
Simulation workflow schema are moved to a plug in https://github.com/nomad-coe/nomad-schema-plugin-simulation-workflow. References to the old schema are removed.


JFRudzinski (approved):
@ladinesa this looks straightforward to me :+1:  thanks for making me aware of it.

PR #80, opened 2024-01-04
added tests for MD method ENUMS in Gromacs

### LINKED ISSUE #61: Add test for sd integrator gromacs

JFRudzinski:
@ladinesa I added the requested tests. These each use only a single configuration of a relatively small system, so they should be efficient. In principle I can make these even smaller, but I will try to address that when I get to #30


JFRudzinski:
@ladinesa So, the new tests are failing in the pipeline even though they all pass for me locally. I did some debugging, and it appears to me that the new tests do not have the correct archive, somehow the data is coming from some other file/test? Very confused, I don't know how this is possible. Any ideas?

JFRudzinski:
@ladinesa not sure if you saw this. I just rebased the branch to see if that would help, but I am still having the same issue: pytests pass locally but the pipeline fails. I am tempted to create a new branch from develop and implement the changes again from scratch since it seems something really strange is going on, but maybe you could take a quick look first and let me know if the tests work locally for you as well?

ladinesa:
> @ladinesa not sure if you saw this. I just rebased the branch to see if that would help, but I am still having the same issue: pytests pass locally but the pipeline fails. I am tempted to create a new branch from develop and implement the changes again from scratch since it seems something really strange is going on, but maybe you could take a quick look first and let me know if the tests work locally for you as well?

@JFRudzinski sorry for the late reply, the problem seems to be that md.log is missing from your commit. This is because *.log is included in .gitignore. Please remove this line and add the md.log files again.

ladinesa:
Are the test files of reasonable size? If indeed I am fine with the tests and you can merge it.

JFRudzinski:
> Are the test files of reasonable size? If indeed I am fine with the tests and you can merge it.

:facepalm: :facepalm: yes, thank you. I remember now I had to do this once before. Can I leave the `*.log` out of the `.gitignore`? It was under the "Django" section.

ladinesa:
> > Are the test files of reasonable size? If indeed I am fine with the tests and you can merge it.
> 
> 🤦 🤦 yes, thank you. I remember now I had to do this once before. Can I leave the `*.log` out of the `.gitignore`? It was under the "Django" section.

I think it would not do any harm. My bad I simply copied this ignore file from nomad I guess because I am lazy hehe. We can also create an ignore file later

JFRudzinski:
> > > Are the test files of reasonable size? If indeed I am fine with the tests and you can merge it.
> > 
> > 
> > 🤦 🤦 yes, thank you. I remember now I had to do this once before. Can I leave the `*.log` out of the `.gitignore`? It was under the "Django" section.
> 
> I think it would not do any harm. My bad I simply copied this ignore file from nomad I guess because I am lazy hehe. We can also create an ignore file later

That's fine. Thanks for the help. Did you want to check this or shall I merge it?

ladinesa:
> > > > Are the test files of reasonable size? If indeed I am fine with the tests and you can merge it.
> > > 
> > > 
> > > 🤦 🤦 yes, thank you. I remember now I had to do this once before. Can I leave the `*.log` out of the `.gitignore`? It was under the "Django" section.
> > 
> > 
> > I think it would not do any harm. My bad I simply copied this ignore file from nomad I guess because I am lazy hehe. We can also create an ignore file later
> 
> That's fine. Thanks for the help. Did you want to check this or shall I merge it?

I am fine with it. my only concern as usual is the file size. if they are manageable i.e. < 1mb then i think everything is fine. Also if the assertions are again wrt to the ones in the output file that would also be great. But these details we can also redo again when we migrate the parsers.

JFRudzinski:
> > > > > Are the test files of reasonable size? If indeed I am fine with the tests and you can merge it.
> > > > 
> > > > 
> > > > 🤦 🤦 yes, thank you. I remember now I had to do this once before. Can I leave the `*.log` out of the `.gitignore`? It was under the "Django" section.
> > > 
> > > 
> > > I think it would not do any harm. My bad I simply copied this ignore file from nomad I guess because I am lazy hehe. We can also create an ignore file later
> > 
> > 
> > That's fine. Thanks for the help. Did you want to check this or shall I merge it?
> 
> I am fine with it. my only concern as usual is the file size. if they are manageable i.e. < 1mb then i think everything is fine. Also if the assertions are again wrt to the ones in the output file that would also be great. But these details we can also redo again when we migrate the parsers.
 
Yes, each test file is ~244k. In principle this could be reduced further, I have opened an issue to reduce the test file sizes in general, but I would like to try to do that together in one pass.

I did think about that, but again I would rather do this all together per parser. I will open issues about this so we don't forget.

ladinesa (approved):

PR #93, opened 2024-02-08
added check to replace non-element atom labels

### LINKED ISSUE #92: Add check for atom labels that are not elements in atomistic parsers
Replace atom labels with 'X' for non-element labels. nb - this task should be moved to the normalizer in the new schema!

JFRudzinski:
@ladinesa added some hacky checks to ensure that atom labels which do not correspond to chemical elements are replaced by `X` to prevent issues in the normalizers. We discussed this in our Area C meeting today, and plan to implement more rigorous support for non-atomic systems within the new schema. Also, this should be checked in the system normalizer. But for now we enforce it in the parsers. I put a TODO there to move this check to the normalizer in the new schema.

Please give my code a check :pray:

### DIFF
--- a/atomisticparsers/h5md/parser.py
+++ b/atomisticparsers/h5md/parser.py
@@ -580,6 +580,15 @@ class H5MDParser(MDParser):
             time = system_info[step].pop("time")
             atoms_dict = system_info[step]
 
+            atom_labels = atoms_dict.get("labels")
+            if atom_labels is not None:
+                try:
+                    from ase import Atoms as ase_atoms
+
+                    _ = ase_atoms(symbols=atom_labels)
+                except Exception:  # TODO this check should be moved to the system normalizer in the new schema
+                    atoms_dict["labels"] = ["X" for _ in atom_labels]
+
             topology = None
             if i_step == 0:  # TODO extend to time-dependent bond lists and topologies
                 atoms_dict["bond_list"] = self._data_parser.get("connectivity.bonds")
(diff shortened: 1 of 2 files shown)


atomisticparsers/h5md/parser.py
```diff
@@ -580,6 +580,15 @@ def parse_system(self):
             time = system_info[step].pop("time")
             atoms_dict = system_info[step]
 
+            atom_labels = atoms_dict.get("labels")
+            if atom_labels is not None:
+                try:
+                    from ase import Atoms as ase_atoms
```
ladinesa: maybe move to general imports

JFRudzinski: done.

atomisticparsers/h5md/parser.py
```diff
@@ -580,6 +580,15 @@ def parse_system(self):
...
+
+                    _ = ase_atoms(symbols=atom_labels)
```
ladinesa: the specific function to test if symbols are valid in ase is 
```python
from ase.symbols import symbols2numbers
try:
    symbols2numbers(labels)
except KeyError:
    labels = ['X'] * len(labels)

ladinesa: but I am also fine with trying to generate the full ase atoms

ladinesa (approved):

atomisticparsers/h5md/parser.py
JFRudzinski: ah, nice, thanks. done.

PR #94, opened 2024-02-26
91 add gromacs support for fe calculations

### LINKED ISSUE #91: Add Gromacs support for FE calculations

JFRudzinski:
@ladinesa I have added support for a new type of calculation in gromacs. These changes go with the schema additions [HERE](https://github.com/nomad-coe/nomad-schema-plugin-simulation-workflow/pull/14). 

In addition to a code review whenever you have a chance, I would appreciate some guidance in terms of how to properly implement testing from now on. Till now I have mostly followed the existing testing architecture, but it seems from our last conversation that we need to diverge from this model. Is there already some good testing examples that I can follow?


atomisticparsers/gromacs/parser.py
```diff
@@ -180,72 +181,101 @@ def str_to_step_info(val_in):
...
+    def __init__(self):
+        super().__init__(None)
 
-        thermostat_tau = input_parameters.get("tau-t", 0) * ureg.ps
+    def init_quantities(self):
+        def str_to_input_parameters(val_in):
+            re_array = re.compile(r"\s*([\w\-]+)\[[\d ]+\]\s*=\s*\{*(.+)")
+            re_scalar = re.compile(r"\s*([\w\-]+)\s*[=:]\s*(.+)")
+            parameters = dict()
+            val = val_in.split("\n")
```
ladinesa: use splitlines()

atomisticparsers/gromacs/parser.py
```diff
@@ -180,72 +181,101 @@ def str_to_step_info(val_in):
...
+            val = [line.strip() for line in val]
```
ladinesa:
```python
val = [line.strip() for line in val_in.splitlines()]

File: atomisticparsers/gromacs/parser.py [hunk 3]
```diff
@@ -180,72 +181,101 @@ def str_to_step_info(val_in):
...
+                "yaxis": "",
+                "column_headers": [],
+            }
+            re_columns = re.compile(r"@\s*s\d{1,2}\s*legend\s*\".*\"")
+            re_title = re.compile(r"@\s*title.*")
+            re_xaxis = re.compile(r"@\s*xaxis.*")
+            re_yaxis = re.compile(r"@\s*yaxis.*")
+            re_comment = re.compile(r"^[@#]")
+            re_quotes = re.compile(r"\"(.*)\"")
+            val = val_in.strip().split("\n")
```
ladinesa: splitlines()

atomisticparsers/gromacs/parser.py
```diff
@@ -180,72 +181,101 @@ def str_to_step_info(val_in):
...
+            val = [line.strip() for line in val]
+            for val_n in val:
+                val_title = re_yaxis.match(val_n)
+                val_xaxis = re_title.match(val_n)
+                val_yaxis = re_xaxis.match(val_n)
+                val_legend = re_columns.match(val_n)
+                val_comment = re_comment.match(val_n)
+                if val_title:
```
ladinesa: for title, xaxis and yaxis, you can combine the regex, and capture the label (title, xaxis, yaxis) which you can use as key in results so you get rid of the other ifs

atomisticparsers/gromacs/parser.py
```diff
@@ -180,72 +181,101 @@ def str_to_step_info(val_in):
...
-        barostat_tau = input_parameters.get("tau-p", 0) * ureg.ps
+        def str_to_results(val_in):
+            results = {
+                "column_vals": None,
+                "title": "",
+                "xaxis": "",
+                "yaxis": "",
+                "column_headers": [],
+            }
+            re_columns = re.compile(r"@\s*s\d{1,2}\s*legend\s*\".*\"")
```
ladinesa: i would put the regex patterns in init so you do not compile everytime

atomisticparsers/gromacs/parser.py
```diff
@@ -640,6 +673,42 @@ def __init__(self):
...
+        """
+        Tries to find the mdp input parameters (ext = mdp) that match the mainfile calculation.
+        Priority is as follows:
+            1. output mdp file containing both the matching mainfile name and the standard
+            gromacs name `mdout`
+            2. file containing the standard gromacs name `mdout`
+            3. input mdp file matching the mainfile name (as usual)
+            4. any `.mdp` file within the directory (as usual)
+        """
+        ext = "mdp"
```
ladinesa: perhaps put ext and std_filename in init? easier to maintain

atomisticparsers/gromacs/parser.py
```diff
@@ -640,6 +673,42 @@ def __init__(self):
...
+            return os.path.join(self._maindir, files[0])
+
+        for f in files:
+            filename = f.rsplit(".", 1)[0]
+            if self._basename in filename and std_filename in filename:
+                return os.path.join(self._maindir, f)
+
+        for f in files:
+            filename = f.rsplit(".", 1)[0]
+            if std_filename in filename:
```
ladinesa: put this in the other loop?

atomisticparsers/gromacs/parser.py
```diff
@@ -1045,10 +1112,184 @@ def parse_method(self):
...
+            else None
+        )
+        thermostat_parameters = {}
+        thermostat_parameters["thermostat_type"] = value
+        if "sd" in integrator:
+            thermostat_parameters["thermostat_type"] = "langevin_goga"
+        if thermostat_parameters["thermostat_type"]:
+            reference_temperature = self.input_parameters.get("ref-t", None)
+            if isinstance(reference_temperature, str):
+                reference_temperature = float(reference_temperature.split()[0])
```
ladinesa: do we need error handling here? it seems you use this quite often so have a small function to deal with this.
```
    def to_float(string):
        try:
            value = float(string)
        except ValueError:
            value = None
        return value

File: atomisticparsers/gromacs/parser.py [hunk 9]
```diff
@@ -1045,10 +1112,184 @@ def parse_method(self):
...
+            barostat_parameters["coupling_constant"] = (
+                np.ones(shape=(3, 3)) * float(taup) * ureg.picosecond if taup else None
+            )
+            refp = self.input_parameters.get("ref-p", None)
+            barostat_parameters["reference_pressure"] = (
+                refp * ureg.bar if refp is not None else None
+            )
+            compressibility = self.input_parameters.get("compressibility", None)
+            barostat_parameters["compressibility"] = (
+                compressibility * (1.0 / ureg.bar)
```
ladinesa, line 1196: you would get error here is compressibility is None

atomisticparsers/gromacs/parser.py
```diff
@@ -1250,8 +1406,80 @@ def parse_workflow(self):
...
                 method["thermodynamic_ensemble"] = "NVE"
 
+            params_key = "free_energy_calculation_parameters"
+            method[params_key] = self.get_free_energy_calculation_parameters()
+
+            self.xvg_parser.mainfile = self.get_gromacs_file("xvg")
+            free_energies = self.xvg_parser.get("results")
+
+            title = free_energies.get("title", "") if free_energies is not None else ""
+            flag_FE = False
```
ladinesa: flag_fe

atomisticparsers/gromacs/parser.py
```diff
@@ -1250,8 +1406,80 @@ def parse_workflow(self):
...
+                    with self.archive.m_context.raw_file(filename, "r+b") as f:
+                        # The expected columns of the xvg file are:
+                        # Total Energy
+                        # dH/dlambda current lambda
+                        # Delta H between each lambda and current lambda (n_lambda columns)
+                        # PV Energy
+                        if columns[:, 2:-1].shape[1] != sec_FE.n_states:
+                            self.logger.warning(
+                                "Unexpected format of xvg file. Not storing free energy calculation results."
+                            )
```
ladinesa: are we still supposed to write if check fails?

ladinesa:
for the changes that you have implemented here, it is reasonable to run the whole parser. For parsers in general, it is hard and not easy mantain if we try to isolate the tests. What we can perhaps do is to have tests only for the individual file parser classes. But again, in the current setup the changes in the file_parser comes with changes in the parser itself. Our previous discussion was about testing schema normalization, but I will need to think deeper how we can have a more robust testing strategy.

JFRudzinski:
> r the individual

Ok, thanks for the insight about the tests, I just wanted to make sure that I do not continue to propagate bad habits. In this case, I will implement the tests as usual with a full example file.

JFRudzinski:
@ladinesa I added some tests here and I have a few questions:
1. I would like to test that my reference between results and method is populated correctly. I tried to do something like
`assert isinstance(sec_results.method_ref, FreeEnergyCalculationParameters())`
But it seems that the schema cannot be imported here, i.e., the following fails:
`from simulationworkflowschema.MolecularDynamics import FreeEnergyCalculationParameters`

Is this approach fundamentally flawed? Or how can I access the schema classes in these tests?

2. How should I test the hdf5 references/storage? It seems that when I run these tests the hdf5 reference is not made (all those references are `None` even though they are populated correctly when I parse the same example myself with the same build. I am guessing it is some issue with the filepath/writing but I don't see any corresponding errors. Any ideas?

atomisticparsers/gromacs/parser.py
JFRudzinski: done.

atomisticparsers/gromacs/parser.py
JFRudzinski: done.

atomisticparsers/gromacs/parser.py
JFRudzinski: done.

atomisticparsers/gromacs/parser.py
JFRudzinski: done.

atomisticparsers/gromacs/parser.py
JFRudzinski: nice, thanks! done.

atomisticparsers/gromacs/parser.py
JFRudzinski: done.

atomisticparsers/gromacs/parser.py
JFRudzinski: I was trying to prioritize the situation where someone incorporated both the basename and std_filename into the file name, e.g., `<basename>_<std_filename>.mdp`. But it is possible that `<basename>_<std_filename>.mdp` and `<std_filename>.mdp` both correspond to files in the directory. I don't see any way to easily use 1 loop for this, do you?

atomisticparsers/gromacs/parser.py
JFRudzinski: done.

ladinesa:
> @ladinesa I added some tests here and I have a few questions:
> 
> 1. I would like to test that my reference between results and method is populated correctly. I tried to do something like
>    `assert isinstance(sec_results.method_ref, FreeEnergyCalculationParameters())`
>    But it seems that the schema cannot be imported here, i.e., the following fails:
>    `from simulationworkflowschema.MolecularDynamics import FreeEnergyCalculationParameters`
> 
> Is this approach fundamentally flawed? Or how can I access the schema classes in these tests?
> 
> 2. How should I test the hdf5 references/storage? It seems that when I run these tests the hdf5 reference is not made (all those references are `None` even though they are populated correctly when I parse the same example myself with the same build. I am guessing it is some issue with the filepath/writing but I don't see any corresponding errors. Any ideas?

1.from simulationworkflowschema.molecular_dynamics import FreeEnergyCalculationParameters

2. They are only created when archive context is present i.e. during peocessing. this i think should be tested in nomad where the processing fixtures are found, otherwise we can also do it in pasrer but more involved. when i am done with the hdf5 ref implementation i will test it.

atomisticparsers/gromacs/parser.py
ladinesa: ah i understand then it is fine.

atomisticparsers/gromacs/parser.py
JFRudzinski: I have a question about the implementation. Often times I am doing some 1 liner operation like `val = float(val) * ureg.nanometer if val else None`.

Is it generally better to do

`val = to_float(val) * ureg.nanometer if to_float(val) else None`
OR
```
val = to_float(val)
val = val * ureg.nanometer if val else None
```

Does your answer depend at all on if val is a local variable or part of the archive, like `workflow.results.final_force_maximum`

ladinesa: you would do to_float twice for 1. so 2 is better

atomisticparsers/gromacs/parser.py
JFRudzinski: no, you are right, I have moved this check above and added a couple others.

atomisticparsers/gromacs/parser.py
JFRudzinski: ok, that's what I thought, just wanted to make sure before implementing. done.

JFRudzinski:
> > @ladinesa I added some tests here and I have a few questions:
> > 
> > 1. I would like to test that my reference between results and method is populated correctly. I tried to do something like
> >    `assert isinstance(sec_results.method_ref, FreeEnergyCalculationParameters())`
> >    But it seems that the schema cannot be imported here, i.e., the following fails:
> >    `from simulationworkflowschema.MolecularDynamics import FreeEnergyCalculationParameters`
> > 
> > Is this approach fundamentally flawed? Or how can I access the schema classes in these tests?
> > 
> > 2. How should I test the hdf5 references/storage? It seems that when I run these tests the hdf5 reference is not made (all those references are `None` even though they are populated correctly when I parse the same example myself with the same build. I am guessing it is some issue with the filepath/writing but I don't see any corresponding errors. Any ideas?
> 
> 1.from simulationworkflowschema.molecular_dynamics import FreeEnergyCalculationParameters
> 
> 2. They are only created when archive context is present i.e. during peocessing. this i think should be tested in nomad where the processing fixtures are found, otherwise we can also do it in pasrer but more involved. when i am done with the hdf5 ref implementation i will test it.

about #1 :facepalm: 

about #2 Ok then I will just add a TODO for now and leave it till we work out the hdf5 functionality a bit more

JFRudzinski:
@ladinesa I am done with my changes. Do you want to check it again or is it safe to merge? (I still need to wait for Chema's review on the schema, and to merge the schema first of course)

PR #98, opened 2024-03-10
Parser interface

ladinesa:
Use parser class hierarchy with common parser method parse and abstract method write_to_archive.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -251,3 +251,4 @@ env/
 
 # gromacs files created on run
 .*.npz
+.*.lock
--- a/atomisticparsers/asap/parser.py
+++ b/atomisticparsers/asap/parser.py
@@ -113,9 +113,9 @@ class AsapParser(MDParser):
         self.parse_method()
 
         # set up md parser
-        self.n_atoms = [
-            traj.get_global_number_of_atoms() for traj in self.traj_parser.traj
-        ]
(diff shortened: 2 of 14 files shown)


JFRudzinski (approved):
Just a general question about parser maintenance: There are a number of these atomsitic parsers which only have a handful (or sometimes 0!) entries + I presume that the functionality of these parsers are not very good / up-to-date. We need to decide which parsers we continue to maintain and which we dump. I think the move to plugins is a reasonable time to make this decision.

Not sure what the best way to go about this other than recent usage. Anyway, not relevant for this MR, but we should discuss sometime.

ladinesa:
> Just a general question about parser maintenance: There are a number of these atomsitic parsers which only have a handful (or sometimes 0!) entries + I presume that the functionality of these parsers are not very good / up-to-date. We need to decide which parsers we continue to maintain and which we dump. I think the move to plugins is a reasonable time to make this decision.
> 
> Not sure what the best way to go about this other than recent usage. Anyway, not relevant for this MR, but we should discuss sometime.

yes, we can certainly do this.

ladinesa:
@JFRudzinski are we ok with this pr?

JFRudzinski:
> @JFRudzinski are we ok with this pr?

Yes please go ahead

PR #99, opened 2024-03-12
Fix nomad dependency

ladinesa:


pyproject.toml
```diff
@@ -3,38 +3,38 @@ requires = ["setuptools"]
 build-backend = "setuptools.build_meta"
 
 [project]
-name = 'atomisticparsers'
-version = '1.0'
-description = 'Collection of NOMAD parsers for atomistic codes.'
+name = "atomisticparsers"
+version = "1.0"
```
JosePizarro3, line 7: What about [dynamic versioning](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html)?

You can check the nomad-analysis pyproject for having an idea: https://github.com/FAIRmat-NFDI/nomad-analysis/blob/77c57770474cf2081e00945d95994f64cd7bf911/pyproject.toml

JosePizarro3 (approved):
Quoting has changed, meaning probably the Ruff settings are different w.r.t. NOMAD

pyproject.toml
ladinesa: We can integrate all these tiny details in the new parsers or in another pr.

ladinesa:
> Member

ruff does not format toml file, i just changed everything to double quotes, for uniformity, but does not really matter.

PR #105, opened 2024-04-19
Update writing of hdf5 data

ladinesa:


JFRudzinski (approved):
Wow, much nicer, thank you @ladinesa!!

@Bernadette-Mohr once Alvin merges this branch, you should rebase your branch and then you can follow this format for the US stuff. When you do that, can you also add tests for these FE calculation HDF5 quantities? I had not done this yet because I was waiting for this new functionality to be implemented.

PR #106, opened 2024-04-25
Fix quantity types

ladinesa:
Fix shape and type inconsistencies due to metainfo refactoring. @TLCFEM

### DIFF
--- a/atomisticparsers/utils/parsers.py
+++ b/atomisticparsers/utils/parsers.py
@@ -210,7 +210,7 @@ class MDParser(Parser):
 
         interactions.sort(key=lambda x: x.get("type"))
         current_type = interactions[0].get("type")
-        interaction_values = {}
+        interaction_values: Dict[str, Any] = {}
         for interaction in interactions:
             interaction_type = interaction.get("type")
             if current_type and current_type != interaction_type:
--- a/atomisticparsers/utils/parsers.py
+++ b/atomisticparsers/utils/parsers.py
@@ -183,98 +183,98 @@ class MDParser(Parser):
             return
 
         ## OLD ##
-        interaction_dict = {}
-        for interaction_key in Interaction.m_def.all_quantities.keys():
-            interaction_dict[interaction_key] = np.array(
-                [interaction.get(interaction_key) for interaction in interactions],
-                dtype=object,
-            )
-        interaction_dict = {key: val for key, val in interaction_dict.items()}
-        interaction_types = (
-            np.unique(interaction_dict["type"])
-            if interaction_dict.get("type") is not None
-            else []
-        )
-        for interaction_type in interaction_types:
-            sec_interaction = Interaction()
-            sec_model.contributions.append(sec_interaction)
-            interaction_indices = np.where(
-                interaction_dict["type"] == interaction_type
-            )[0]
-            sec_interaction.type = interaction_type
-            sec_interaction.n_interactions = len(interaction_indices)
-            sec_interaction.n_atoms
-            for key, val in interaction_dict.items():
-                if key == "type":
-                    continue
-                interaction_vals = val[interaction_indices]
-                if type(interaction_vals[0]).__name__ == "ndarray":
-                    interaction_vals = np.array(
-                        [vals.tolist() for vals in interaction_vals], dtype=object
-                    )
-                if interaction_vals.all() is None:
-                    continue
-                if key == "parameters":
-                    interaction_vals = interaction_vals.tolist()
-                elif key == "n_atoms":
-                    interaction_vals = interaction_vals[0]
-                if hasattr(sec_interaction, key):
-                    sec_interaction.m_set(
-                        sec_interaction.m_get_quantity_definition(key), interaction_vals
-                    )
-
-            if not sec_interaction.n_atoms:
-                sec_interaction.n_atoms = (
-                    len(sec_interaction.get("atom_indices")[0])
-                    if sec_interaction.get("atom_indices") is not None
-                    else None
-                )
-        ## NEW ##
-        # def write_interaction_values(values):
+        # interaction_dict = {}
+        # for interaction_key in Interaction.m_def.all_quantities.keys():
+        #     interaction_dict[interaction_key] = np.array(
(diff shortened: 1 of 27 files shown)


atomisticparsers/utils/parsers.py
```diff
@@ -179,52 +179,53 @@ def parse_md_workflow(self, data: Dict[str, Any]) -> None:
...
-        interaction_types = (
-            np.unique(interaction_dict["type"])
-            if interaction_dict.get("type") is not None
-            else []
-        )
-        for interaction_type in interaction_types:
+        if not interactions:
+            return
+
+        def write_interaction_values(values):
```
JFRudzinski, line 185: In general this function is now a bit confusing to me,  I think I do not yet fully understand all your motivations here. Please clarfiy.

atomisticparsers/utils/parsers.py
```diff
@@ -179,52 +179,53 @@ def parse_md_workflow(self, data: Dict[str, Any]) -> None:
...
-            interaction_indices = np.where(
-                interaction_dict["type"] == interaction_type
-            )[0]
-            sec_interaction.type = interaction_type
-            sec_interaction.n_interactions = len(interaction_indices)
-            sec_interaction.n_atoms
-            for key, val in interaction_dict.items():
+            sec_interaction.type = current_type
+            sec_interaction.n_atoms = max(
+                [len(v) for v in values.get("atom_indices", [[0]])]
```
JFRudzinski, line 190: Should we really give a default of [[0]] if there are no atom_indices instead of just not populating any interactions sections? Can this even occur?

atomisticparsers/utils/parsers.py
```diff
@@ -179,52 +179,53 @@ def parse_md_workflow(self, data: Dict[str, Any]) -> None:
...
+            )
+            for key, val in values.items():
+                # TODO tempory fix: atom_labels, atom_indices not homogeneous
+                # fill in missing atom label with 'X', atom index with -1
```
JFRudzinski: I don't think it is actually what the function below is doing, and I don't understand what you mean by fill in missing values? I guess you are thinking a case where atom_indices and atom_labels are of a different length than the actual list of atoms? This would be a very strange occurrence from my perspective and should result in an error, unless we have a concrete use case for this.

atomisticparsers/utils/parsers.py
```diff
@@ -179,52 +179,53 @@ def parse_md_workflow(self, data: Dict[str, Any]) -> None:
...
+                if key in ["atom_indices", "atom_labels"]:
+                    val = [
+                        (
+                            v
```
JFRudzinski: This is the reason for the errors in the tests dealing with interactions and the bond list that you have commented out. In the case of "atom_indices" you are subtracting 1 from the atom_index in all cases, which results in a -1-indexed list of indices, and all the test assertions to be off by 1.

atomisticparsers/utils/parsers.py
```diff
@@ -179,52 +179,53 @@ def parse_md_workflow(self, data: Dict[str, Any]) -> None:
...
+                            + [-1 if key == "atom_indices" else "X"]
```
JFRudzinski: In line 133/4 of utils/mdanalysis.py I have made a check for atom_names and replaced the labels by X if they do not correspond to actual elements. We should employ a similar logic in general here, but not based on individual missing labels, as mentioned above. If there are no atom label info, then we can replace all by X, otherwise we should check the names/labels individually to see if they correspond to elements. Again, not sure if there exists a case where the atom_label/name list does not match the length of system.atoms. Please clarify if I am wrong.

JFRudzinski (commented):
Most changes look perfectly fine. I would like to fix the issues in the interaction section before your merge this.

atomisticparsers/utils/parsers.py
ladinesa: I made this part simpler to get rid of the mapping of the data to the interaction types, now I simply sort the interactions by type and collect the values until a new interaction type is encountered then write the interaction values. I had a temporary fix for the atom_indices and atom_labels because before, the shape (arbitrary) and type (np.object) is not consistent with the definition hence the error. The more formal fix, I think is to restructure the defintion so we get we are not compelled to have a fixed array.

atomisticparsers/utils/parsers.py
ladinesa: From the denition atom_labels and atom_indices have shape ['n_interactions', 'n_atoms'] but n_atoms is not the same for all interactions. Is this right? Hence the fix that I have here that fulfills the definition is to set n_atoms to the max value  l189 and for interactions with n-atoms less than max value, I fill in 'X' or -1 respectively. As I have mentioned above, this is not an ideal fix and it is better to redefine Interaction section so we do not group all interactions.

atomisticparsers/utils/parsers.py
ladinesa: i really do not like this fix from me and it is simply wrong. We need to fix the rewrite the definition instead.

atomisticparsers/utils/parsers.py
ladinesa: Yes sure I can put a return here.

atomisticparsers/utils/parsers.py
JFRudzinski: To go back to a flat list, I need to check and adjust my implementation, because this is used for things like the bond list. How do you want to proceed?

ladinesa: We can perhaps just go for the temporary fix for now, because you are also working on a new schema which probably will not have this complication anyway.

JFRudzinski:
@ladinesa I think there are some major misunderstandings happening here. I would like to come to a consensus and fix the issues (i.e., I don't want to wait for the new schema), but would like to chat first. Can you let me know when you are available?

ladinesa:
@JFRudzinski can we merge thus now?

JFRudzinski:
> @JFRudzinski can we merge thus now?

Yes, sorry, I thought that was assumed, please go ahead.

PR #107, opened 2024-05-15
Fix interactions

JFRudzinski:


JFRudzinski:
@ladinesa I did not find any case with inhomogeneous interactions. I simply removed your fix and everything works fine.

In fe_test, there are interactions that don't have atom_indices or atom_labels at all, but that should be fine, they are just describing the general force field.

Can you have a look again? If you find an example with inhomogeneous interactions just point me to it so I can take a look and try to come up with a fix.

JFRudzinski:
@ladinesa I made a fix in parse_interactions() in mdanalysis.py that avoids this lammps situation with the inhomogeneous atom_labels. I still did not see a case with inhomogeneous indices. Hopefully this is enough for now. 

You will also notice that I remove some storage of (incorrect) parameters. When we work on the new schema and parser plugins, we will do a better job grabbing some of the interaction parameter information by parsing the raw files directly (MDAnalysis does not robustly access this info)

Let me know what you think and if you see any other potential issues.

ladinesa:
Hi @JFRudzinski I just run the tests and I did not find any more errors. Please merge.

PR #111, opened 2024-06-19
Support for energy per mole in H5MD

### LINKED ISSUE #108: Add support for mole in energy quantities in H5MD parser

JFRudzinski:
Added support for use of energy/mole e.g., KJ/mole as units for energy quantities

### DIFF
--- a/atomisticparsers/h5md/parser.py
+++ b/atomisticparsers/h5md/parser.py
@@ -536,11 +536,20 @@ class H5MDParser(MDParser):
                 if (
                     "energ" in observable_name
                 ):  # TODO check for energies or energy when matching name
-                    if hasattr(Energy, observable_label):
-                        data["energy"][observable_label] = dict(value=val)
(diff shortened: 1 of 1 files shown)


JFRudzinski:
@Bernadette-Mohr I just made some small changes to support usage of energy per mole quantities with the H5MD parser. Take a look and let me know if you have any suggestions or questions.

Also, I thought this would be a good opportunity to walk you through rebasing just to make sure we are on the same page, since it is our standard way for updating changes that have been made to develop after we have already checkout out a branch.

So, go ahead and create your branch for the small adjustment to the Gromacs parser, and then let me know when you have done that. I will then merge this branch (after your review), and then we can meet to do the rebase, ok?

Bernadette-Mohr (commented):
I guess the amount of substance will most likely be given in mole, we probably don't need to cover any more esoteric options. I don't see any issues.

PR #113, opened 2024-06-19
removed deprecated dependency on archive.m_context

### LINKED ISSUE #112: update gromacs parser

Bernadette-Mohr:


JFRudzinski:
@ladinesa Can you take a look at this? The installation is suddenly failing in the pipeline. It was working fine when I made a PR last night. The changes in this PR have no connection with this, but just to be sure I made a simple copy of develop and just added a comment to the code, and get the same errors.

ladinesa:
I think it is because mpcdf is down. maybe we can rerun later.

JFRudzinski:
> I think it is because mpcdf is down. maybe we can rerun later.

oh ok, makes sense...I did not notice this, thanks

JFRudzinski:
@Bernadette-Mohr It's fixed! You can go ahead and merge it now

Issue #122, opened 2024-08-23
Gromacs processing fails due to incorrect parsing of input params


JFRudzinski:
Example Upload: https://nomad-lab.eu/prod/v1/staging/gui/user/uploads/upload/id/_oYhRQxiTdiECWVQAiZu8A

error shows `input_parameters.rlist` is empty

JFRudzinski:
fixed by #125

PR #125, opened 2024-09-03
122 rebased to entry point

JFRudzinski:
rebased branch for #122


JFRudzinski:
@ladinesa just a small question: If I run `nomad parse` locally, something with the writing of HDF5Datasets is breaking. Is this something that you know about or is there a way around this?

ladinesa:
> @ladinesa just a small question: If I run `nomad parse` locally, something with the writing of HDF5Datasets is breaking. Is this something that you know about or is there a way around this?

this should be expected as creating an h5 ref needs a server context mraning it is only possible to do during processing. currently metainfo raises an error but i think it should obly be a warning. The workaround is to put if archive.m_context

JFRudzinski:
> > @ladinesa just a small question: If I run `nomad parse` locally, something with the writing of HDF5Datasets is breaking. Is this something that you know about or is there a way around this?
> 
> this should be expected as creating an h5 ref needs a server context mraning it is only possible to do during processing. currently metainfo raises an error but i think it should obly be a warning. The workaround is to put if archive.m_context

ok great, this makes sense. Definitely agree it should be a warning so that one can still run the parser in this way. Would you like me to make an issue for this on gitlab or will you remember?

ladinesa:
> > > @ladinesa just a small question: If I run `nomad parse` locally, something with the writing of HDF5Datasets is breaking. Is this something that you know about or is there a way around this?
> > 
> > 
> > this should be expected as creating an h5 ref needs a server context mraning it is only possible to do during processing. currently metainfo raises an error but i think it should obly be a warning. The workaround is to put if archive.m_context
> 
> ok great, this makes sense. Definitely agree it should be a warning so that one can still run the parser in this way. Would you like me to make an issue for this on gitlab or will you remember?

i@TLCFEM  ehat do you thibk

atomisticparsers/gromacs/parser.py
```diff
@@ -312,30 +376,39 @@ class GromacsMDAnalysisParser(MDAnalysisParser):
...
+            self.logger.warning(
+                'Reading force field from tpr not yet supported for Gromacs 2024. Interactions will not be stored'
+            )
+            return []
 
         with open(self.mainfile, 'rb') as f:
             data = tpr_utils.TPXUnpacker(f.read())
 
-        interactions = []
+        interactions: List[Any] = []
```
JFRudzinski: @ladinesa not sure about this, I got an error in the tests that told me to annotate this line...?

ladinesa: probably you have to be more specific? This is a list of tuples right?

atomisticparsers/h5md/parser.py
```diff
@@ -890,6 +890,7 @@ def write_to_archive(self) -> None:
         self.parse_observable_info()
         self.parse_parameter_info()
 
+        print('in parsing')
```
ladinesa: remove pls

atomisticparsers/gromacs/parser.py
```diff
@@ -80,24 +83,85 @@ def str_to_header(val_in):
             return {v[0].strip(): v[1].strip() for v in val if len(v) == 2}
 
         def str_to_input_parameters(val_in):
-            re_array = re.compile(r'\s*([\w\-]+)\[[\d ]+\]\s*=\s*\{*(.+)')
+            re_section = re.compile(r'^\s*([\w\-]+):\s*$')
```
ladinesa: maybe put these variables in the init method so it is not compiled everytime

atomisticparsers/gromacs/parser.py
```diff
@@ -80,24 +83,85 @@ def str_to_header(val_in):
...
+                val_n = val_n.rstrip()  # Remove trailing spaces
+                if not val_n:
+                    continue
+
+                current_indent = len(val_n) - len(
+                    val_n.lstrip()
+                )  # Calculate the indentation level
+
+                # Handle end of section based on indentation
+                while indent_levels and current_indent <= indent_levels[-1]:
```
ladinesa, line 108: is there a case where this does not break? i suggest to reformulate this using for loop or can we not do with previous_indent instead of indent_levels since we are sort of identifying the level wrt to previous?

atomisticparsers/gromacs/parser.py
```diff
@@ -80,24 +83,85 @@ def str_to_header(val_in):
...
+                    stack.pop()
+                    indent_levels.pop()
+
+                # Check for section
```
ladinesa, line 112: can you make the check for section and sub-section use a common loop?

atomisticparsers/gromacs/parser.py
```diff
@@ -80,24 +83,85 @@ def str_to_header(val_in):
...
-            re_array = re.compile(r'\s*([\w\-]+)\[[\d ]+\]\s*=\s*\{*(.+)')
+            re_section = re.compile(r'^\s*([\w\-]+):\s*$')
+            re_subsection = re.compile(r'^\s*([\w\-]+\s[\d]+):\s*$')
             re_scalar = re.compile(r'\s*([\w\-]+)\s*[=:]\s*(.+)')
+            re_array = re.compile(r'\s*([\w\-]+)\[[\d ]+\]\s*=\s*\{*(.+)')
+            re_shorthand_array = re.compile(
+                r'\s*([\w\-]+)\[\d+,\.\.\.,\d+\]\s*=\s*\{(\d+),\.\.\.,(\d+)\}'
+            )
+
             parameters = dict()
```
ladinesa, line 94: not sure why parameters is defined

atomisticparsers/gromacs/parser.py
```diff
@@ -80,24 +83,85 @@ def str_to_header(val_in):
...
+                    indent_levels.append(current_indent)
+                    continue
+
+                # Check for scalar
                 val_scalar = re_scalar.match(val_n)
                 if val_scalar:
-                    parameters[val_scalar.group(1)] = val_scalar.group(2)
+                    key = val_scalar.group(1)
+                    value = val_scalar.group(2)
+                    if value.lower() == 'true':
```
ladinesa: a bit picky but u can condense: if value.lower() in ['true', 'false']: value = value.lower() == 'true'

atomisticparsers/gromacs/parser.py
```diff
@@ -80,24 +83,85 @@ def str_to_header(val_in):
...
+                        value = True
+                    elif value.lower() == 'false':
+                        value = False
+                    elif value.replace('.', '', 1).isdigit():
```
ladinesa, line 137: you can also use mod operator to check if int

atomisticparsers/gromacs/parser.py
```diff
@@ -312,30 +376,39 @@ class GromacsMDAnalysisParser(MDAnalysisParser):
...
-    def get_interactions(self):
+    def get_interactions(self, gromacs_version: str = None):
         interactions = super().get_interactions()
 
         # add force field parameters
-        try:
-            interactions.extend(self.get_force_field_parameters())
-        except Exception:
-            self.logger.error('Error parsing force field parameters.')
+        # try:
```
ladinesa: do we still need these?

atomisticparsers/gromacs/parser.py
```diff
@@ -1457,6 +1566,39 @@ def parse_workflow(self):
                 sec_fe.value_total_energy_differences_magnitude = columns[:, 2:-1]
                 sec_fe.value_PV_energy_magnitude = columns[:, -1]
 
+    def standardize_input_parameters_dict_recursive(self, input_dict: dict):
```
ladinesa: maybe drop _dict_recursive in the name

ladinesa (approved):

JFRudzinski:
@ladinesa Let me try to explain a bit the motivation for the regex changes here: For more advanced functionalities, the gromacs log file stores the input parameters in a hierarchical structure that groups certain quantities, and also introduces sub-quantities. Since the name of these sub-quantities can repeat, we need to keep the hierarchy in our extracted `input_parameters`, e.g.,

```
   pull-group 0:
     atom: not available
     weight: not available
     pbcatom                        = -1
   pull-group 1:
     atom (35):
        atom[0,...,34] = {0,...,34}
     weight: not available
     pbcatom                        = 1280
   pull-group 2:
     atom (1):
        atom[0]=7370
     weight: not available
     pbcatom                        = -1
```

Here, in addition to some of these not being read properly in the old regex function, `atom` would be overwritten several times

Overall, you can see the end results by looking at the [test input log](https://github.com/nomad-coe/atomistic-parsers/blob/e2893044b0432b46819ba391d8cae9b5eec48af4/tests/data/gromacs/input_parameters/test.log) and resulting [input_parameters](https://github.com/nomad-coe/atomistic-parsers/blob/e2893044b0432b46819ba391d8cae9b5eec48af4/tests/data/gromacs/input_parameters/input_parameters.json)

JFRudzinski:
Now, there is still an issue with the current implemented function as seen in this [log](https://raw.githubusercontent.com/nomad-coe/atomistic-parsers/122-rebased-to-entry-point/tests/data/gromacs/free_energy_calculations/alchemical_transformation_single_run/fep_run-7.log)

```
   n-lambdas                      = 11
   separate-dvdl:
       fep-lambdas =   FALSE
      mass-lambdas =   FALSE
      coul-lambdas =   FALSE
       vdw-lambdas =   TRUE
    bonded-lambdas =   FALSE
 restraint-lambdas =   FALSE
temperature-lambdas =   FALSE
all-lambdas:
       fep-lambdas =            0           0           0           0           0           0           0           0           0           0           0
      mass-lambdas =            0           0           0           0           0           0           0           0           0           0           0
      coul-lambdas =            0           0           0           0           0           0           0           0           0           0           0
       vdw-lambdas =            0         0.1         0.2         0.3         0.4         0.5         0.6         0.7         0.8         0.9           1
    bonded-lambdas =            0           0           0           0           0           0           0           0           0           0           0
 restraint-lambdas =            0           0           0           0           0           0           0           0           0           0           0
temperature-lambdas =            0           0           0           0           0           0           0           0           0           0           0
   calc-lambda-neighbors          = -1
   dhdl-print-energy              = total
   sc-alpha                       = 0.5
   sc-power                       = 1
   sc-r-power                     = 6
   sc-sigma                       = 0.3
```

The groupings in the function are based on the indentation, but stupidly the sub-quantities under `all-lambas` and `separate-dvdl` try to align the `=`, such that the indentation hierarchy is lost for `temperature-lambdas` and so this is stored at the root level and also over-written by the second instance

JFRudzinski:
oops, I broke something with the regex function...

ladinesa:
> @ladinesa Let me try to explain a bit the motivation for the regex changes here: For more advanced functionalities, the gromacs log file stores the input parameters in a hierarchical structure that groups certain quantities, and also introduces sub-quantities. Since the name of these sub-quantities can repeat, we need to keep the hierarchy in our extracted `input_parameters`, e.g.,
> 
> ```
>    pull-group 0:
>      atom: not available
>      weight: not available
>      pbcatom                        = -1
>    pull-group 1:
>      atom (35):
>         atom[0,...,34] = {0,...,34}
>      weight: not available
>      pbcatom                        = 1280
>    pull-group 2:
>      atom (1):
>         atom[0]=7370
>      weight: not available
>      pbcatom                        = -1
> ```
> 
> Here, in addition to some of these not being read properly in the old regex function, `atom` would be overwritten several times
> 
> Overall, you can see the end results by looking at the [test input log](https://github.com/nomad-coe/atomistic-parsers/blob/e2893044b0432b46819ba391d8cae9b5eec48af4/tests/data/gromacs/input_parameters/test.log) and resulting [input_parameters](https://github.com/nomad-coe/atomistic-parsers/blob/e2893044b0432b46819ba391d8cae9b5eec48af4/tests/data/gromacs/input_parameters/input_parameters.json)

(can't we simply rely on the key then not on the indent?), sorry i understand now.  we need the indent for new sub section

atomisticparsers/gromacs/parser.py
JFRudzinski: Did you mean like `elif value % 1 == 0:`, this didn't work for me

JFRudzinski:
> > @ladinesa Let me try to explain a bit the motivation for the regex changes here: For more advanced functionalities, the gromacs log file stores the input parameters in a hierarchical structure that groups certain quantities, and also introduces sub-quantities. Since the name of these sub-quantities can repeat, we need to keep the hierarchy in our extracted `input_parameters`, e.g.,
> > ```
> >    pull-group 0:
> >      atom: not available
> >      weight: not available
> >      pbcatom                        = -1
> >    pull-group 1:
> >      atom (35):
> >         atom[0,...,34] = {0,...,34}
> >      weight: not available
> >      pbcatom                        = 1280
> >    pull-group 2:
> >      atom (1):
> >         atom[0]=7370
> >      weight: not available
> >      pbcatom                        = -1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Here, in addition to some of these not being read properly in the old regex function, `atom` would be overwritten several times
> > Overall, you can see the end results by looking at the [test input log](https://github.com/nomad-coe/atomistic-parsers/blob/e2893044b0432b46819ba391d8cae9b5eec48af4/tests/data/gromacs/input_parameters/test.log) and resulting [input_parameters](https://github.com/nomad-coe/atomistic-parsers/blob/e2893044b0432b46819ba391d8cae9b5eec48af4/tests/data/gromacs/input_parameters/input_parameters.json)
> 
> can't we simply rely on the key then not on the indent?

You mean asking to grab known keys/sections I guess? I guess I wanted to write a completely general parsing because it's difficult to anticipate or keep up to date on all the options here and I wanted to just dump this dictionary into the archive for possible exploration by someone. Maybe it's not necessary, let's discuss...

JFRudzinski:
> oops, I broke something with the regex function...

ok fixed again now

JFRudzinski:
@ladinesa here are 2 log examples for the input parameter parsing, one with "pull" options and one with "free energy" options

[solute_in_bilayer_production-1.9.log](https://github.com/user-attachments/files/16906019/solute_in_bilayer_production-1.9.log)
[fep_run-7.log](https://github.com/user-attachments/files/16906030/fep_run-7.log)

ladinesa:
> > > @ladinesa Let me try to explain a bit the motivation for the regex changes here: For more advanced functionalities, the gromacs log file stores the input parameters in a hierarchical structure that groups certain quantities, and also introduces sub-quantities. Since the name of these sub-quantities can repeat, we need to keep the hierarchy in our extracted `input_parameters`, e.g.,
> > > ```
> > >    pull-group 0:
> > >      atom: not available
> > >      weight: not available
> > >      pbcatom                        = -1
> > >    pull-group 1:
> > >      atom (35):
> > >         atom[0,...,34] = {0,...,34}
> > >      weight: not available
> > >      pbcatom                        = 1280
> > >    pull-group 2:
> > >      atom (1):
> > >         atom[0]=7370
> > >      weight: not available
> > >      pbcatom                        = -1
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Here, in addition to some of these not being read properly in the old regex function, `atom` would be overwritten several times
> > > Overall, you can see the end results by looking at the [test input log](https://github.com/nomad-coe/atomistic-parsers/blob/e2893044b0432b46819ba391d8cae9b5eec48af4/tests/data/gromacs/input_parameters/test.log) and resulting [input_parameters](https://github.com/nomad-coe/atomistic-parsers/blob/e2893044b0432b46819ba391d8cae9b5eec48af4/tests/data/gromacs/input_parameters/input_parameters.json)
> > 
> > 
> > can't we simply rely on the key then not on the indent?
> 
> You mean asking to grab known keys/sections I guess? I guess I wanted to write a completely general parsing because it's difficult to anticipate or keep up to date on all the options here and I wanted to just dump this dictionary into the archive for possible exploration by someone. Maybe it's not necessary, let's discuss...

Your approach actually makes total sense, I just made a little cosmetic improvement to enable the addition of regex and converter easier:

```python
re_section = re.compile(r'(?P<indent> +?)(?P<key>\S+?) +\(*(?P<index>\d*)\)*\:')
re_value = re.compile(r'(?P<indent> +?)(?P<key>\S+?) *[:=]+ *(?P<value>[^\n]+)')

converters = [
    (re.compile(r'([-+]?\d+)'), lambda x: int(x.group(1))),
    (re.compile(r'(?i:(?:true|false))'), lambda x: x.group(1).lower() == 'true'),
    (re.compile(r'([-+]?\d+\.*\d*(?:[Ee][-+]\d+)?)'), lambda x: float(x.group(1))),
    (re.compile(r'\{(\d+),\.\.\.,(\d+)\}'), lambda x: list(range(int(x.group(1)), int(x.group(2))))),
    (re.compile(r'\{*([\d\.]+,.+)\}*'), lambda x: [float(v) for v in x.group(1).split(',')])
]

def convert_value(value):
    value = value.strip()
    for pattern, converter in converters:
        match = pattern.match(value)
        if match:
            return converter(match)
    if value.lower() == 'not available':
        return None
    return value

def parse_indented(block):
    root = {}
    stack = [(root, -1)]

    for line in block.splitlines():
        match = re_section.match(line)
        if not match:
            match = re_value.match(line)
        if not match:
            continue

        match_dct = match.groupdict()
        indent = len(match_dct.get('indent'))
        key = match_dct.get('key')
        value = match_dct.get('value')

        while stack and stack[-1][1] >= indent:
            stack.pop()

        parent = stack[-1][0]

        child = convert_value(value )if value else {}
        if key in parent:
            if isinstance(parent[key], list):
                parent[key].append(child)
            else:
                parent[key] = [parent[key], child]
        else:
            parent[key] = child

        stack.append((child, indent))
    return root
```

JFRudzinski:
@ladinesa Thanks a lot for taking a look at this code and for your suggestions, I really do appreciate it

I am, however, having quite a few issues trying to get this code to work:

1. just a small change in case you test it, the first converter has no group(1), it should be:
```python
            converters = [
                (re.compile(r'([-+]?\d+)'), lambda x: int(x.group(1))),
                (
                    re.compile(r'(?i:(?:true|false))'),
                    lambda x: x.group().lower() == 'true',
                ), ...
```

2. array groups like
```
   posres-com (3):
      posres-com[0]= 0.00000e+00
      posres-com[1]= 0.00000e+00
      posres-com[2]= 0.00000e+00
```
are no longer treated correctly but are being stored like `'posres-com': {'posres-com[0]': 0, 'posres-com[1]': 0, 'posres-com[2]': 0},` in the stack or possibly sometimes not stored at all.

3. Groups like `qm-opts:` or `all-lambdas:` with no value are not being stored at all (they should be stored as a key with a dict value with everyone at deeper indent inside the dict

4. In the attached test log, the code currently crashes after `n-lambdas` (I can look into this myself but I don't want to continue until we address the other points)
[fep_run-7.log](https://github.com/user-attachments/files/16963092/fep_run-7.log)

JFRudzinski:
@ladinesa how would you feel about keeping my changes and then improving this function further in a later PR? As I mentioned before I am under some pressure to get this specific use-case working and published in some fashion within the next week or 2. 

If you think the code is too risky then of course we will continue to deal with it, but I am concerned that neither of us have that much time to fix it atm

ladinesa:
> @ladinesa how would you feel about keeping my changes and then improving this function further in a later PR? As I mentioned before I am under some pressure to get this specific use-case working and published in some fashion within the next week or 2.
> 
> If you think the code is too risky then of course we will continue to deal with it, but I am concerned that neither of us have that much time to fix it atm

we could also do that, your current approach is already good actually.

JFRudzinski:
ok thanks, then I will transfer you suggestion and any other info to an issue and then clean up things here

PR #126, opened 2024-09-09
Adapt changes to atomutils

ladinesa:


ladinesa:
@JFRudzinski , i can also merge this now no? it is just changing the imports from simulation workflow schema

PR #131, opened 2024-09-13
115 problematic parsing of traj files from ase

### LINKED ISSUE #115: problematic parsing of *.traj files from ase
Something seems to have been changed between 2022 and February 2023. 
Earlier *.traj seemed to have been parsed correctly, i.e.at least the structure was extracted. See e.g. [here](https://nomad-lab.eu/prod/v1/gui/search/entries/entry/id/zv3A2M3O8crylImLxuCeSWHQWnvy), for the same file the structure is not extracted currently. 
Since last year, even though also the asap parser recognised the file, the structure was not extracted, see .e.g. [here](https://nomad-lab.eu/prod/v1/gui/search/entries/entry/id/jUyRN5InXCBel0YhSTRD2XPkwShc)

ladinesa:


schumannj:
I have tested this branch with my trajectory files and it works well in getting the structure as well as energies and forces. 

One potential issue I noticed, is that the name of the calculator, (in my case "vasp") is listed under method>force_field>model>
Screenshot 2024-11-11 171618

Not sure it there is a clear way to tell if it was originally from a force field calculation or DFT. The traj.files I have available from my own previous work also have a range of level of detail. So it would be difficult to extract something general.
For my own use case it is enough to have structures and energies for now.

JFRudzinski:
merged with #140

PR #145, opened 2025-02-06
139 visualize systems with topology information only

### LINKED ISSUE #139: Visualize systems with topology information only
Currently, systems are only visualized if a trajectory is part of the uploaded data. Since we only display a single snapshot, providing a coordinate file, e.g., `gro` or `pdb`, should be enough for nomad to create the visualization.

Bernadette-Mohr:
Added visualization option from a single frame (`*.gro` file or `*.pdb` file).
Parser now ensures an `*.edr` file is present in the upload before attempting to access it, preventing an `AttributeError`.

### DIFF
--- a/atomisticparsers/gromacs/parser.py
+++ b/atomisticparsers/gromacs/parser.py
@@ -820,9 +820,9 @@ class GromacsParser(MDParser):
         edr_file = self.get_gromacs_file('edr')
         if edr_file:
             # If the edr file can't be read or there are no keys, an error will be raised by the energy parser
+            self.energy_parser.mainfile = edr_file
             try:
                 self.energy_parser.keys()
-                self.energy_parser.mainfile = edr_file
                 thermo_data = self.energy_parser
             except Exception:
                 thermo_data = None
@@ -849,6 +849,9 @@ class GromacsParser(MDParser):
         self.thermodynamics_steps = [
             int(time / time_step if time_step else 1) for time in calculation_times
         ]
+        if not self.thermodynamics_steps:
+            self.logger.error('No thermodynamic data can be found.')
+            return
 
         for n, step in enumerate(self.thermodynamics_steps):
             data = {
@@ -1647,7 +1650,7 @@ class GromacsParser(MDParser):
 
             return primary_file
 
-        # Enumerate over copy of list to avoid skipping elements
+        # Iterate over copy of list to avoid skipping elements
         for ext in self.traj_file_priority_list[:]:
             traj_file = _get_traj_file(ext)
             if traj_file:
--- a/atomisticparsers/gromacs/parser.py
+++ b/atomisticparsers/gromacs/parser.py
@@ -1695,7 +1695,9 @@ class GromacsParser(MDParser):
         """
         traj_priority_list = ['trr', 'xtc', 'pdb', 'gro']
 
-        def is_readable(file_path):
+        # Check if the number of particles in the current trajectory file matches
+        # the number of particles in self.mainfile (topology).
+        def matches_mainfile(file_path):
             positions = None
             try:
                 test_universe = MDAnalysis.Universe(
@@ -1713,18 +1715,20 @@ class GromacsParser(MDParser):
 
             return False
 
-        # sort all entries in self.gromacs_files by priority_list and filename match
+        # Sort all entries in self.gromacs_files by priority_list and filename match,
+        # return matching trajectory with highest priority.
         fallback_files = []
         for file_ext in traj_priority_list:
             results = self.get_all_gromacs_files(file_ext)
             if results:
                 traj_files_list = [*results.get(0, []), *results.get(1, [])]
                 for file_path in traj_files_list:
-                    if is_readable(file_path):
+                    if matches_mainfile(file_path):
                         return [file_path]
                 fallback_files.extend(results.get(2, []))
+        # return the first match from fallback_files only if no higher-priority match was found
         for file_path in fallback_files:
-            if is_readable(file_path):
+            if matches_mainfile(file_path):
                 return [file_path]
         return []
 
--- a/atomisticparsers/gromacs/parser.py
+++ b/atomisticparsers/gromacs/parser.py
@@ -774,50 +774,11 @@ class GromacsParser(MDParser):
 
         return self.get_gromacs_file(self.mdp_ext)
 
-    def get_gromacs_file(self, ext, return_all=False):
+    def get_gromacs_file(self, ext):
         files = [d for d in self._gromacs_files if d.endswith(ext)]
 
         if len(files) == 0:
             return ''
-
-        if return_all:
-            all_files = {}
-            contain_basename, no_basename_match, counts = [], [], []
-            for f in files:
-                # We assume that the matching trajectory file has (or starts with)
-                # the same basename as the log file, e.g., out.log would correspond to
-                # out.tpr and out.trr and out.edr (or out_some.tpr ...).
-                # These files are therefore given first (second... ) priority in the dictionary.
-                if f.rsplit('.', 1)[0] == self._basename:
-                    all_files[0] = {0: os.path.join(self._maindir, f)}
-                elif f.rsplit('.', 1)[0].startswith(self._basename):
-                    contain_basename.append(os.path.join(self._maindir, f))
-                else:
-                    # if potential files are named differently, we guess that the ones with a unique name
-                    # would be the files we are more likely interested in, and sort the results accordingly
-                    count = 0
-                    for reff in self._gromacs_files:
-                        if f.rsplit('.', 1)[0] == reff.rsplit('.', 1)[0]:
-                            count += 1
-                    counts.append(count)
-                    no_basename_match.append(os.path.join(self._maindir, f))
-            # add the files starting with the same basename in the order they were encountered
-            if contain_basename:
-                all_files[1] = {
-                    idx: file_path for idx, file_path in enumerate(contain_basename)
-                }
-            # sort files with the correct format, but different basename:
-            # according to the number of other files with the same name
-            # (fewer files with same name: higher priority)
-            if no_basename_match:
-                all_files[2] = {
-                    idx: file_path
-                    for idx, file_path in enumerate(
-                        [x for _, x in sorted(zip(counts, no_basename_match))]
-                    )
-                }
-
-            return all_files
 
         if len(files) == 1:
             return os.path.join(self._maindir, files[0])
@@ -844,13 +805,48 @@ class GromacsParser(MDParser):
                 if f.rsplit('.', 1)[0] == reff.rsplit('.', 1)[0]:
                     count += 1
             if count == 1:
-                if return_all:
-                    all_files.extend(os.path.join(self._maindir, f))
-                else:
-                    return os.path.join(self._maindir, f)
+                return os.path.join(self._maindir, f)
             counts.append(count)
 
         return os.path.join(self._maindir, files[counts.index(min(counts))])
+
+    def get_all_gromacs_files(self, ext):
+        files = [d for d in self._gromacs_files if d.endswith(ext)]
+
+        if len(files) == 0:
+            return ''
+
+        all_files = {}
+        contain_basename, no_basename_match, counts = [], [], []
+        for f in files:
+            # We assume that the matching trajectory file has (or starts with)
+            # the same basename as the log file, e.g., out.log would correspond to
+            # out.tpr and out.trr and out.edr (or out_some.tpr ...).
+            # These files are therefore given first (second... ) priority in the dictionary.
+            if f.rsplit('.', 1)[0] == self._basename:
+                all_files[0] = [os.path.join(self._maindir, f)]
+            # ? `a in b` or `b.startswith(a) or b.endswith(a)`? How generic/specific do we want to be?
+            elif self._basename in f.rsplit('.', 1)[0]:
+                contain_basename.append(os.path.join(self._maindir, f))
+            else:
+                # if potential files are named differently, we guess that the ones with a unique name
+                # would be the files we are more likely interested in, and sort the results accordingly
+                count = 0
+                for reff in self._gromacs_files:
+                    if f.rsplit('.', 1)[0] == reff.rsplit('.', 1)[0]:
+                        count += 1
+                counts.append(count)
+                no_basename_match.append(os.path.join(self._maindir, f))
+        # add the files starting with the same basename in the order they were encountered
+        if contain_basename:
+            all_files[1] = contain_basename
+        # sort files with the correct format, but different basename:
+        # according to the number of other files with the same name
+        # (fewer files with same name: higher priority)
+        if no_basename_match:
+            all_files[2] = [x for _, x in sorted(zip(counts, no_basename_match))]
+
+        return all_files
 
     def parse_thermodynamic_data(self):
         sec_run = self.archive.run[-1]
@@ -1685,8 +1681,13 @@ class GromacsParser(MDParser):
     def find_trajectory_files(self):
         """
         Find and set trajectory files following the priority:
-        "trr" > "xtc", fall back "pdb" > "gro".
-        Prioritize files with exact basename match over files with partial basename match over other files.
+        "trr" > "xtc", fallback "pdb" > "gro". Prioritize files with exact basename
+        match over files with partial basename match with the same extension.
+        Test other files only if no true trajectory format with exact or partial
+        basename match is found.
+
+        Example: out.log -> out.trr > out_1.trr > 2_out.trr > out.xtc > ... > out.pdb > out.gro
+
         """
         traj_priority_list = ['trr', 'xtc', 'pdb', 'gro']
 
@@ -1708,48 +1709,20 @@ class GromacsParser(MDParser):
 
             return False
 
-        def sort_by_priority(files_dicts):
-            # accounting for multiple files starting with the basename (priority 1)
-            # we expect each dict to contain the exact basename match (priority 0)
-            n_keys = [len(d[1]) if d.get(1) else 0 for d in files_dicts]
-            m_keys = [len(d[2]) if d.get(2) else 0 for d in files_dicts]
-            seen_keys = []
-            dicts_list = []
-            n_add = 1
-            m_add = sum(n_keys) + 1
-            for d_idx, d in enumerate(files_dicts):
-                for key in d:
-                    if key == 1:
-                        d[key] = {idx + n_add: val for idx, val in d[key].items()}
-                        n_add += n_keys[d_idx]
-                    if key == 2:
-                        d[key] = {idx + m_add: val for idx, val in d[key].items()}
-                        m_add += m_keys[d_idx]
-                    dicts_list.append(d.get(key, None))
-                    for idx in d[key].keys():
-                        if idx not in seen_keys:
-                            seen_keys.append(idx)
-            # make sure the returned list is sorted by our priority order
-            sorted_values = [
-                d.get(key, None) for key in sorted(seen_keys) for d in dicts_list
-            ]
-            # remove None values from not present files
-            sorted_values = list(filter(lambda x: x is not None, sorted_values))
-
-            return sorted_values
-
         # sort all entries in self.gromacs_files by priority_list and filename match
-        traj_files_list = []
-        for p in traj_priority_list:
-            results = self.get_gromacs_file(p, return_all=True)
-            # if files with the correct extension are found, add to the list
+        fallback_files = []
+        for file_ext in traj_priority_list:
+            results = self.get_all_gromacs_files(file_ext)
             if results:
-                traj_files_list.append(results)
-        if traj_files_list:
-            traj_files_list = sort_by_priority(traj_files_list)
-            for file_path in traj_files_list:
-                if is_readable(file_path):
-                    return [file_path]
+                traj_files_list = [*results.get(0, []), *results.get(1, [])]
+                for file_path in traj_files_list:
+                    if is_readable(file_path):
+                        print(file_path)
+                        return [file_path]
+                fallback_files.extend(results.get(2, []))
+        for file_path in fallback_files:
+            if is_readable(file_path):
+                return [file_path]
         return []
 
     def write_to_archive(self):
--- a/atomisticparsers/gromacs/parser.py
+++ b/atomisticparsers/gromacs/parser.py
@@ -363,6 +363,8 @@ class GromacsEDRParser(FileParser):
         self._results[key] = val
 
     def keys(self):
+        if self.fileedr is None:
+            return []
         return list(self.fileedr.keys())
 
     @property
@@ -860,19 +862,10 @@ class GromacsParser(MDParser):
         # TODO read also from ene
         # Make sure *.edr file is part of upload before attempting to parse it.
         edr_file = self.get_gromacs_file('edr')
-        thermo_data = None
         if edr_file:
-            # If the edr file can't be read or there are no keys, an error will be raised by the energy parser
+            self.energy_parser.keys()
             self.energy_parser.mainfile = edr_file
-            try:
-                self.energy_parser.keys()
-                thermo_data = self.energy_parser
-            except Exception:
-                self.logger.error('Error reading edr file.')
-            # self.energy_parser.mainfile = edr_file
-            # self.energy_parser.keys()
-            # thermo_data = self.energy_parser
-
+            thermo_data = self.energy_parser
         if not thermo_data:
             # try to get it from log file
             steps = self.input_parameters.get('step', [])
--- a/atomisticparsers/gromacs/parser.py
+++ b/atomisticparsers/gromacs/parser.py
@@ -776,45 +776,9 @@ class GromacsParser(MDParser):
             if self.mdp_std_filename in filename:
                 return os.path.join(self._maindir, f)
 
-        return next(iter(self.get_all_gromacs_files(self.mdp_ext)), None)
+        return next(iter(self.get_gromacs_files(self.mdp_ext)), None)
 
-    def get_gromacs_file(self, ext):
-        files = [d for d in self._gromacs_files if d.endswith(ext)]
-
-        if len(files) == 0:
-            return ''
-
-        if len(files) == 1:
-            return os.path.join(self._maindir, files[0])
-
-        # we assume that the file has the same basename as the log file e.g.
-        # out.log would correspond to out.tpr and out.trr and out.edr
-        for f in files:
-            if f.rsplit('.', 1)[0] == self._basename:
-                return os.path.join(self._maindir, f)
-
-        for f in files:
-            if f.rsplit('.', 1)[0].startswith(self._basename):
-                return os.path.join(self._maindir, f)
-
-        # if the files are all named differently, we guess that the one that does not
-        # share the same basename would be file we are interested in
-        # e.g. in a list of files out.log someout.log out.tpr out.trr another.tpr file.trr
-        # we guess that the out.* files belong together and the rest that does not share
-        # a basename would be grouped together
-        counts = []
-        for f in files:
-            count = 0
-            for reff in self._gromacs_files:
-                if f.rsplit('.', 1)[0] == reff.rsplit('.', 1)[0]:
-                    count += 1
-            if count == 1:
-                return os.path.join(self._maindir, f)
-            counts.append(count)
-
-        return os.path.join(self._maindir, files[counts.index(min(counts))])
-
-    def get_all_gromacs_files(
+    def get_gromacs_files(
         self, ext, return_all=False
     ) -> Union[Tuple[List[str], List[str], List[str]], List[str]]:
         """
@@ -887,7 +851,7 @@ class GromacsParser(MDParser):
 
         # TODO read also from ene
         # Make sure *.edr file is part of upload before attempting to parse it.
-        edr_file = next(iter(self.get_all_gromacs_files('edr')), None)
+        edr_file = next(iter(self.get_gromacs_files('edr')), None)
         if edr_file:
             self.energy_parser.keys()
             self.energy_parser.mainfile = edr_file
@@ -1169,7 +1133,7 @@ class GromacsParser(MDParser):
         try:
             n_atoms = self.traj_parser.get('n_atoms', 0)
         except Exception:
-            gro_file = next(iter(self.get_all_gromacs_files('gro')), None)
+            gro_file = next(iter(self.get_gromacs_files('gro')), None)
             self.traj_parser.mainfile = gro_file
             n_atoms = self.traj_parser.get('n_atoms', 0)
         atoms_info = self.traj_parser.get('atoms_info', {})
@@ -1578,9 +1542,7 @@ class GromacsParser(MDParser):
             params_key = 'free_energy_calculation_parameters'
             method[params_key] = self.get_free_energy_calculation_parameters()
 
-            self.xvg_parser.mainfile = next(
-                iter(self.get_all_gromacs_files('xvg')), None
-            )
+            self.xvg_parser.mainfile = next(iter(self.get_gromacs_files('xvg')), None)
             free_energies = self.xvg_parser.get('results')
 
             title = free_energies.get('title', '') if free_energies is not None else ''
@@ -1739,7 +1701,7 @@ class GromacsParser(MDParser):
             contains_files: List[str] = []
             fallback_files: List[str] = []
             for file_ext in traj_priority_group:
-                traj_files_tup = self.get_all_gromacs_files(file_ext, return_all=True)
+                traj_files_tup = self.get_gromacs_files(file_ext, return_all=True)
                 if traj_files_tup:
                     for file_path in traj_files_tup[0]:
                         if matches_mainfile(file_path):
@@ -1784,7 +1746,7 @@ class GromacsParser(MDParser):
         self._frame_rate = None
 
         header = self.log_parser.get('header', {})
-        topology_file = next(iter(self.get_all_gromacs_files('tpr')), None)
+        topology_file = next(iter(self.get_gromacs_files('tpr')), None)
         sec_run = Run()
         sec_run.program = Program(
             name='GROMACS',
--- a/atomisticparsers/gromacs/parser.py
+++ b/atomisticparsers/gromacs/parser.py
@@ -17,6 +17,7 @@
 # limitations under the License.
 #
 import os
+import sys
 import numpy as np
 import logging
 import re
@@ -774,8 +775,37 @@ class GromacsParser(MDParser):
 
         return self.get_gromacs_file(self.mdp_ext)
 
-    def get_gromacs_file(self, ext):
+    def get_gromacs_file(self, ext, return_all=False):
         files = [d for d in self._gromacs_files if d.endswith(ext)]
+
+        if return_all:
+            all_files = {}
+            no_basename_match, counts = [], []
+            for f in files:
+                # we assume that the file has the same basename as the log file e.g.
+                # out.log would correspond to out.tpr and out.trr and out.edr
+                if f.rsplit('.', 1)[0] == self._basename:
+                    all_files[0] = os.path.join(self._maindir, f)
+                elif f.rsplit('.', 1)[0].startswith(self._basename):
+                    all_files[1] = os.path.join(self._maindir, f)
+                else:
+                    # if the files are all named differently, we guess that the one that does not
+                    # share the same basename would be file we are more likely interested in and sort
+                    # the results accordingly
+                    count = 0
+                    for reff in self._gromacs_files:
+                        if f.rsplit('.', 1)[0] == reff.rsplit('.', 1)[0]:
+                            count += 1
+                    counts.append(count)
+                    no_basename_match.append(os.path.join(self._maindir, f))
+            if no_basename_match:
+                for idx, val in enumerate(
+                    [x for _, x in sorted(zip(counts, no_basename_match))]
+                ):
+                    all_files[idx + 2] = val
+            sorted(all_files.items())
+
+            return all_files
 
         if len(files) == 0:
             return ''
@@ -805,7 +835,10 @@ class GromacsParser(MDParser):
                 if f.rsplit('.', 1)[0] == reff.rsplit('.', 1)[0]:
                     count += 1
             if count == 1:
-                return os.path.join(self._maindir, f)
+                if return_all:
+                    all_files.extend(os.path.join(self._maindir, f))
+                else:
+                    return os.path.join(self._maindir, f)
             counts.append(count)
 
         return os.path.join(self._maindir, files[counts.index(min(counts))])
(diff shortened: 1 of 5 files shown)


atomisticparsers/gromacs/parser.py
```diff
@@ -840,7 +841,10 @@ def parse_thermodynamic_data(self):
 
         if not thermo_data:
             # get it from edr file
-            thermo_data = self.energy_parser
+            try:
+                thermo_data = self.energy_parser
+            except Exception as e:
+                self.logger.warning(f'Error parsing edr file: {e}')
```
ladinesa: no return here?

ladinesa: log msg should not be formatted string, remove e, add it as exc_info

atomisticparsers/gromacs/parser.py
```diff
@@ -1617,8 +1621,11 @@ def parse_input(self):
         trajectory_file = os.path.basename(self.traj_parser.auxilliary_files[0])
         sec_input_output_files.x_gromacs_inout_file_trajtrr = trajectory_file
 
-        edr_file = os.path.basename(self.energy_parser.mainfile)
-        sec_input_output_files.x_gromacs_inout_file_eneredr = edr_file
+        try:
+            edr_file = os.path.basename(self.energy_parser.mainfile)
+            sec_input_output_files.x_gromacs_inout_file_eneredr = edr_file
+        except TypeError:
+            logging.warning(f'Error parsing *.edr file, no energy data available.')
```
ladinesa: remove string format

atomisticparsers/gromacs/parser.py
```diff
@@ -1629,6 +1636,48 @@ def parse_input(self):
...
+        Find and set trajectory files following the priority:
+        "trr" > "xtc" > "pdb" > "gro".
+        """
+
+        def _get_file_or_fallback(primary_ext, secondary_ext):
+            """Helper function to get primary file, or fallback to secondary."""
+            primary_file = self.get_gromacs_file(primary_ext)
+            primary_file_nopath = primary_file.rsplit('.', 1)[0]
+            primary_file_nopath = primary_file_nopath.rsplit('/')[-1]
+            secondary_file = self.get_gromacs_file(secondary_ext)
```
ladinesa: execute only secondary file search if primary file is not found.

atomisticparsers/gromacs/parser.py
```diff
@@ -1629,6 +1636,48 @@ def parse_input(self):
...
+
+            if trajectory_file:
+                return [trajectory_file]
+            else:
+                # Try pdb first, fallback to gro
+                trajectory_file = _get_file_or_fallback('pdb', 'gro')
+                if trajectory_file:
+                    return [trajectory_file]
+        except FileNotFoundError:
+            logging.warning(f'No coordinates found, no visualization possible.')
```
ladinesa: return [] here so you do not need additional handling

atomisticparsers/gromacs/parser.py
```diff
@@ -1629,6 +1636,48 @@ def parse_input(self):
...
+            trajectory_file = _get_file_or_fallback('trr', 'xtc')
+
+            if trajectory_file:
+                return [trajectory_file]
+            else:
+                # Try pdb first, fallback to gro
+                trajectory_file = _get_file_or_fallback('pdb', 'gro')
+                if trajectory_file:
+                    return [trajectory_file]
+        except FileNotFoundError:
```
ladinesa: are there other exceptions?

atomisticparsers/gromacs/parser.py
```diff
@@ -840,7 +841,10 @@ def parse_thermodynamic_data(self):
 
         if not thermo_data:
```
JFRudzinski: Isn't this completely redundant? We already checked if the edr files has keys to parse

atomisticparsers/gromacs/parser.py
```diff
@@ -816,12 +816,13 @@ def parse_thermodynamic_data(self):
...
-
-        # get it from edr file
-        if self.energy_parser.keys():
-            thermo_data = self.energy_parser
+        if edr_file:
+            self.energy_parser.mainfile = edr_file
+            # get it from edr file
+            if self.energy_parser.keys():
+                thermo_data = self.energy_parser
         else:
```
JFRudzinski: What if you have an edr file but there are no keys? Then you won't end up in here! 

```python
if edr_file:
    if keys:

if not thermo_data:
    <try getting from input>
```

right?

atomisticparsers/gromacs/parser.py
```diff
@@ -1629,6 +1636,48 @@ def parse_input(self):
...
+                    secondary_file
+                    if secondary_file_nopath.startswith(self._basename)
+                    else primary_file
+                )
+            else:
+                trajectory_file = primary_file
+
+            return trajectory_file
+
+        try:
```
JFRudzinski: I don't really understand a few things here:

1. Why the `try:`, where can it break in error exactly?

2. Why separate the trr/xtc vs. pdb/gro cases? Can't you do something like:

```python
def find_trajectory_files(self):
        def _get_traj_file(ext):
            primary_file = self.get_gromacs_file(primary_ext)
            if primary_file is '':
                return None
            primary_file_nopath = primary_file.rsplit('.', 1)[0]  #! Check what happens if there are multiple files found in get_gromacs_file
            primary_file_nopath = primary_file_nopath.rsplit('/')[-1]
            return primary_file_nopath

    traj_file_priority_list = ['trr', 'xtc', 'pdb', 'gro'] 
    for ext in traj_file_priority_list:
        traj_file = _get_traj_file(ext)
        if traj_file:
            return [traj_file]

    return []
```

atomisticparsers/gromacs/parser.py
Bernadette-Mohr: This got restructured, now every check has an output that triggers correct behavior later on.

atomisticparsers/gromacs/parser.py
Bernadette-Mohr: Thanks, I wasn't aware of this.

atomisticparsers/gromacs/parser.py
Bernadette-Mohr: The logic in this function should be more efficient now.

atomisticparsers/gromacs/parser.py
```diff
@@ -816,13 +816,17 @@ def parse_thermodynamic_data(self):
...
-        self.energy_parser.mainfile = edr_file
-
-        # get it from edr file
-        if self.energy_parser.keys():
-            thermo_data = self.energy_parser
-        else:
+        if edr_file:
+            # If the edr file can't be read or there are no keys, an error will be raised by the energy parser
+            self.energy_parser.mainfile = edr_file
+            try:
```
JFRudzinski: Do we still need the try? I think if the edr file exists then the calling of keys() should work even if there is none. Is there some other case you found where this breaks?

Also, if there are no keys then you will not parse anything, but in that case you should try to parse the log file.

atomisticparsers/gromacs/parser.py
```diff
@@ -1617,8 +1620,11 @@ def parse_input(self):
         trajectory_file = os.path.basename(self.traj_parser.auxilliary_files[0])
         sec_input_output_files.x_gromacs_inout_file_trajtrr = trajectory_file
 
-        edr_file = os.path.basename(self.energy_parser.mainfile)
-        sec_input_output_files.x_gromacs_inout_file_eneredr = edr_file
+        try:
```
JFRudzinski: Can you explain under what conditions you expect this to break?

atomisticparsers/gromacs/parser.py
```diff
@@ -1690,30 +1722,62 @@ def write_to_archive(self):
...
-        self.traj_parser.auxilliary_files = [trajectory_file]
-
-        # check to see if the trr file can be read properly (and has positions), otherwise try xtc file instead
-        positions = None
-        if (universe := self.traj_parser.universe) is not None:
-            atoms = getattr(universe, 'atoms', None)
-            positions = getattr(atoms, 'positions', None)
-        if positions is None:
-            self.traj_parser.auxilliary_files = [xtc_file] if xtc_file else [trr_file]
+        # JFR comment: I have no idea if output trajectory file can be specified in input
```
JFRudzinski: Please remove these comments, I do not know what I meant, it's probably moot at this point

atomisticparsers/gromacs/parser.py
```diff
@@ -1690,30 +1722,62 @@ def write_to_archive(self):
...
+                self.logger.warning(
+                    'No positions in trajectory file: {}'.format(
+                        self.traj_parser.auxilliary_files[0].split('/')[-1]
+                    )
+                )
+
+                # Check if other files with the same extension are available, attempt to generate universe from them
+                failed_ext = self.traj_parser.auxilliary_files[0].split('.')[-1]
+                failed_file = self.traj_parser.auxilliary_files[0].split('/')[-1]
+                _files = [f for f in self._gromacs_files if f.endswith(failed_ext)]
```
JFRudzinski: This is sort of bypassing the priority logic already implemented in get_gromacs_file().

One route would be to amend get_gromacs_file() to return a list of all files of an ext in the order of priority and then loop through this list for each ext in order

The question is if there is a .trr without any sort of naming match to the log, but an .xtc with an exact naming match, should we use the .xtc? In this case, we should construct a single global priority list for all extensions according to this type of logic and then simply loop through this, breaking when the universe works

atomisticparsers/gromacs/parser.py
```diff
@@ -1690,30 +1722,62 @@ def write_to_archive(self):
...
-            positions = getattr(atoms, 'positions', None)
-        if positions is None:
-            self.traj_parser.auxilliary_files = [xtc_file] if xtc_file else [trr_file]
+        # JFR comment: I have no idea if output trajectory file can be specified in input
+        # ? Do you mean, e.g., a flag in the *.log file that specifies the output trajectory format?
+        self.traj_file_priority_list = ['trr', 'xtc', 'pdb', 'gro']
+        self.traj_parser.auxilliary_files = self.find_trajectory_files()
+        if self.traj_parser.auxilliary_files:
+            # Iterate through available trajectory formats until a valid one is found
+            while self.traj_file_priority_list:
```
JFRudzinski: As indicated below, I would make this a for loop over some sort of global priority list of files.

atomisticparsers/gromacs/parser.py
```diff
@@ -1690,30 +1722,62 @@ def write_to_archive(self):
...
-        positions = None
-        if (universe := self.traj_parser.universe) is not None:
-            atoms = getattr(universe, 'atoms', None)
-            positions = getattr(atoms, 'positions', None)
-        if positions is None:
-            self.traj_parser.auxilliary_files = [xtc_file] if xtc_file else [trr_file]
+        # JFR comment: I have no idea if output trajectory file can be specified in input
+        # ? Do you mean, e.g., a flag in the *.log file that specifies the output trajectory format?
+        self.traj_file_priority_list = ['trr', 'xtc', 'pdb', 'gro']
+        self.traj_parser.auxilliary_files = self.find_trajectory_files()
```
JFRudzinski: I don't really think this function is necessary per se. As stated below, I would adjust get_gromacs_files() to return a prioritized list for each extension. Then this function should construct a global prioritization to be looped over below

JFRudzinski (commented):
@Bernadette-Mohr It's looking good, I think we can improve the traj file logic a bit, and maybe streamline the coding a little. See my comments and then we could discuss this afternoon if anything is unclear

atomisticparsers/gromacs/parser.py
Bernadette-Mohr: This does break when `edr_file` is `None` (no edr file is part of the upload). `TypeError: expected str, bytes or os.PathLike object, not NoneType`, (traces back to nomad-FAIR/nomad/parsing/parsers.py, line 217, `parser.parse(mainfile_path, entry_archive, logger=logger, **kwargs)`).

atomisticparsers/gromacs/parser.py
Bernadette-Mohr: Good point, I'll think about it.

atomisticparsers/gromacs/parser.py
Bernadette-Mohr: I'm not sure how relevant this is. But if the `edr` file is present, but can't be read, this will crash without the try.
`AttributeError: 'NoneType' object has no attribute 'keys'` (line 366, `return list(self.fileedr.keys())`).

JFRudzinski: Ok I also don't know if or under what conditions this is possible. I feel like in that case you might actually want the code to break, or else it won't be apparent that reading the file is an issue...but let me see the final code first and think about it

atomisticparsers/gromacs/parser.py
```diff
@@ -1629,6 +1682,76 @@ def parse_input(self):
...
+                d.get(key, None) for key in sorted(seen_keys) for d in dicts_list
+            ]
+            # remove None values from not present files
+            sorted_values = list(filter(lambda x: x is not None, sorted_values))
+
+            return sorted_values
+
+        # sort all entries in self.gromacs_files by priority_list and filename match
+        traj_files_list = []
+        for p in traj_priority_list:
```
JFRudzinski: Be more explicit about what p is: e.g., `p` --> `file_ext`

atomisticparsers/gromacs/parser.py
```diff
@@ -774,12 +774,51 @@ def get_mdp_file(self):
...
 
+        if return_all:
+            all_files = {}
+            contain_basename, no_basename_match, counts = [], [], []
+            for f in files:
+                # We assume that the matching trajectory file has (or starts with)
+                # the same basename as the log file, e.g., out.log would correspond to
+                # out.tpr and out.trr and out.edr (or out_some.tpr ...).
+                # These files are therefore given first (second... ) priority in the dictionary.
+                if f.rsplit('.', 1)[0] == self._basename:
```
JFRudzinski: Can you add helper functions for each of these cases?
 - The first 2 just return os.path.join
 - The last one does the counting logic

This way you can minimize repeated code and also the code readability will be better, cause it will be more clear that below you do the same thing but don't concatenate them

atomisticparsers/gromacs/parser.py
```diff
@@ -774,12 +774,51 @@ def get_mdp_file(self):
 
         return self.get_gromacs_file(self.mdp_ext)
 
-    def get_gromacs_file(self, ext):
+    def get_gromacs_file(self, ext, return_all=False):
```
JFRudzinski: I don't love that this function is returning completely different data structures depending on what it finds. I know that it was like that before, but there are not so many instances calling this function, I think we can standardize the return and then properly type it. Maybe we just discuss further when we meet...

atomisticparsers/gromacs/parser.py
```diff
@@ -774,12 +774,51 @@ def get_mdp_file(self):
...
+            contain_basename, no_basename_match, counts = [], [], []
+            for f in files:
+                # We assume that the matching trajectory file has (or starts with)
+                # the same basename as the log file, e.g., out.log would correspond to
+                # out.tpr and out.trr and out.edr (or out_some.tpr ...).
+                # These files are therefore given first (second... ) priority in the dictionary.
+                if f.rsplit('.', 1)[0] == self._basename:
+                    all_files[0] = {0: os.path.join(self._maindir, f)}
+                elif f.rsplit('.', 1)[0].startswith(self._basename):
+                    contain_basename.append(os.path.join(self._maindir, f))
```
JFRudzinski: shouldn't this list be called `startswith_basename`?

atomisticparsers/gromacs/parser.py
```diff
@@ -774,12 +774,51 @@ def get_mdp_file(self):
...
+                    # if potential files are named differently, we guess that the ones with a unique name
+                    # would be the files we are more likely interested in, and sort the results accordingly
+                    count = 0
+                    for reff in self._gromacs_files:
+                        if f.rsplit('.', 1)[0] == reff.rsplit('.', 1)[0]:
+                            count += 1
+                    counts.append(count)
+                    no_basename_match.append(os.path.join(self._maindir, f))
+            # add the files starting with the same basename in the order they were encountered
+            if contain_basename:
```
JFRudzinski: why not just create this dict as you go? ...also not quite sure yet why you are using a dict

atomisticparsers/gromacs/parser.py
```diff
@@ -774,12 +774,51 @@ def get_mdp_file(self):
...
+                all_files[1] = {
+                    idx: file_path for idx, file_path in enumerate(contain_basename)
+                }
+            # sort files with the correct format, but different basename:
+            # according to the number of other files with the same name
+            # (fewer files with same name: higher priority)
+            if no_basename_match:
```
JFRudzinski: same here, why a dict?

atomisticparsers/gromacs/parser.py
```diff
@@ -1629,6 +1682,76 @@ def parse_input(self):
...
+                self.logger.warning(
+                    'Error reading positions from trajectory file: {}'.format(file_path)
+                )
+            if positions is not None:
+                # If readable, the correct trajectory has been found
+                return True
+
+            return False
+
+        def sort_by_priority(files_dicts):
```
JFRudzinski: I have the feeling that this function might be simplified but I need to walk through it with you when we meet

JFRudzinski: I think I have a slightly more streamlined approach, but my main question is what is our approach for ordering type `2` files with no basename match if there are multiple options and also multiple extensions?

JFRudzinski (commented):
@Bernadette-Mohr It's looking good. I think just some minor changes in the approach can help simplify it. Take a look at my comments first, then we can meet to walk through it and discuss further.

atomisticparsers/gromacs/parser.py
```diff
@@ -810,19 +810,66 @@ def get_gromacs_file(self, ext):
...
+        if len(files) == 0:
+            return ''
+
+        all_files = {}
+        contain_basename, no_basename_match, counts = [], [], []
+        for f in files:
+            # We assume that the matching trajectory file has (or starts with)
+            # the same basename as the log file, e.g., out.log would correspond to
+            # out.tpr and out.trr and out.edr (or out_some.tpr ...).
+            # These files are therefore given first (second... ) priority in the dictionary.
```
JFRudzinski: can you clean these comments a bit, they are a bit verbose and more confusing than anything. I think you just need to say something like "sort files with .ext into 3 priority categories" at the top, then briefly describe each case directly above the if or elif statement

atomisticparsers/gromacs/parser.py
```diff
@@ -810,19 +810,66 @@ def get_gromacs_file(self, ext):
...
+            if f.rsplit('.', 1)[0] == self._basename:
+                all_files[0] = [os.path.join(self._maindir, f)]
+            # ? `a in b` or `b.startswith(a) or b.endswith(a)`? How generic/specific do we want to be?
+            elif self._basename in f.rsplit('.', 1)[0]:
+                contain_basename.append(os.path.join(self._maindir, f))
```
JFRudzinski: just directly append to all_files[1]

atomisticparsers/gromacs/parser.py
```diff
@@ -810,19 +810,66 @@ def get_gromacs_file(self, ext):
...
+        all_files = {}
+        contain_basename, no_basename_match, counts = [], [], []
+        for f in files:
+            # We assume that the matching trajectory file has (or starts with)
+            # the same basename as the log file, e.g., out.log would correspond to
+            # out.tpr and out.trr and out.edr (or out_some.tpr ...).
+            # These files are therefore given first (second... ) priority in the dictionary.
+            if f.rsplit('.', 1)[0] == self._basename:
+                all_files[0] = [os.path.join(self._maindir, f)]
+            # ? `a in b` or `b.startswith(a) or b.endswith(a)`? How generic/specific do we want to be?
```
JFRudzinski: I think it's not bad to sort these with a sub-priority. In that case I would do starts with > ends with > in

you could effectively copy what you do with the counts except just assign 0, 1, 2 for the 3 cases into  a variable like count for the other case, then you can use the same sorting logic

atomisticparsers/gromacs/parser.py
JFRudzinski: unless you sort them as mentioned above, then add the sorting variable and leave contain_basename and sort below

atomisticparsers/gromacs/parser.py
```diff
@@ -1629,6 +1678,53 @@ def parse_input(self):
...
 
+    def find_trajectory_files(self):
+        """
+        Find and set trajectory files following the priority:
+        "trr" > "xtc", fallback "pdb" > "gro". Prioritize files with exact basename
+        match over files with partial basename match with the same extension.
+        Test other files only if no true trajectory format with exact or partial
+        basename match is found.
+
+        Example: out.log -> out.trr > out_1.trr > 2_out.trr > out.xtc > ... > out.pdb > out.gro
```
JFRudzinski: Example - For out.log: out.trr > out_1.trr

Just the -> is too close to >

Add the additional sorting explanation if added to type 1 above

atomisticparsers/gromacs/parser.py
JFRudzinski: ok, I guess it's fine then

JFRudzinski (commented):
Great! :raised_hands:  Just make the last few changes I suggested, then get a review from Alvin and ask about the tests

atomisticparsers/gromacs/parser.py
```diff
@@ -810,19 +810,66 @@ def get_gromacs_file(self, ext):
 
         return os.path.join(self._maindir, files[counts.index(min(counts))])
 
+    def get_all_gromacs_files(self, ext):
```
ladinesa: I suggest to extend the current `get_gromacs_file` function to take a bool arg `check_if_readable`.

atomisticparsers/gromacs/parser.py
```diff
@@ -1629,6 +1678,53 @@ def parse_input(self):
...
+
+        # sort all entries in self.gromacs_files by priority_list and filename match
+        fallback_files = []
+        for file_ext in traj_priority_list:
+            results = self.get_all_gromacs_files(file_ext)
+            if results:
+                traj_files_list = [*results.get(0, []), *results.get(1, [])]
+                for file_path in traj_files_list:
+                    if is_readable(file_path):
+                        print(file_path)
```
ladinesa: remove

atomisticparsers/gromacs/parser.py
ladinesa: a better fix would be to insert in l366 `if self.fileedr is None: return []`

atomisticparsers/gromacs/parser.py
ladinesa: yes, please also add this logic when you extend `get_gromacs_file`,

atomisticparsers/gromacs/parser.py
```diff
@@ -1695,7 +1695,9 @@ def find_trajectory_files(self):
         """
         traj_priority_list = ['trr', 'xtc', 'pdb', 'gro']
 
-        def is_readable(file_path):
+        # Check if the number of particles in the current trajectory file matches
+        # the number of particles in self.mainfile (topology).
+        def matches_mainfile(file_path):
```
Bernadette-Mohr: I have changed the unfortunate naming here to make it more obvious what I want to achieve.

I think we should discuss whether we want a generic check to determine whether a file is readable for the old `get_gromacs_file` function (or even more generalized) and whether there is a good way to consolidate the not-quite-identical functionalities of `get_gromacs_file` and `get_all_gromacs_files`.

atomisticparsers/gromacs/parser.py
```diff
@@ -810,19 +810,70 @@ def get_gromacs_file(self, ext):
 
         return os.path.join(self._maindir, files[counts.index(min(counts))])
 
+    def get_all_gromacs_files(self, ext):
+        files = [d for d in self._gromacs_files if d.endswith(ext)]
+
+        if len(files) == 0:
+            return None
+
+        all_files = {}
```
ladinesa: not sure why it returns a dict, and the values later concatenated

atomisticparsers/gromacs/parser.py
ladinesa: i think you can replace the former with the latter

atomisticparsers/gromacs/parser.py
```diff
@@ -1629,6 +1682,56 @@ def parse_input(self):
...
+
+        # Sort all entries in self.gromacs_files by priority_list and filename match,
+        # return matching trajectory with highest priority.
+        fallback_files = []
+        for file_ext in traj_priority_list:
+            results = self.get_all_gromacs_files(file_ext)
+            if results:
+                traj_files_list = [*results.get(0, []), *results.get(1, [])]
+                for file_path in traj_files_list:
+                    if matches_mainfile(file_path):
```
ladinesa: match_mainfile is expensive, is there no other way to determine if the traj file is valid without loading it?

Bernadette-Mohr: I replaced the contents of `matches_mainfile` with just attempting to set `self.traj_parser.auxilliary_files` with the current trajectory file path, passing on any error from the `MDAnalysisParser.universe`. This at least doesn't add another universe creation attempt.
Given the freedom with GROMACS output file naming, I don't see a way to prevent occasionally having to make more than one such attempt to find a matching trajectory.

JFRudzinski (commented):
I am not sure if this review request is new or old, but please clear this MR with @ladinesa before I do a final check.

atomisticparsers/gromacs/parser.py
Bernadette-Mohr: I followed the suggestion by @ladinesa , it seems to handle my edge case, the `try` is removed.

atomisticparsers/gromacs/parser.py
Bernadette-Mohr: I replaced `get_gromacs_file` by `get_all_gromacs_files` for every occurrence in the parser. It does add a small overhead because of the different output data structures. An alternative would be to relax the prioritization by filename similarity within `find_trajectory_file` and make `get_all_gromacs_files` simply return a list.

atomisticparsers/gromacs/parser.py
Bernadette-Mohr: `get_gromacs_file` is now tentatively replaced by `get_all_gromacs_files`. If everybody agrees, I can remove the old function.

atomisticparsers/gromacs/parser.py
```diff
@@ -810,19 +822,64 @@ def get_gromacs_file(self, ext):
 
         return os.path.join(self._maindir, files[counts.index(min(counts))])
 
+    def get_all_gromacs_files(self, ext):
+        files = [d for d in self._gromacs_files if d.endswith(ext)]
+
+        if len(files) == 0:
+            return
+
+        all_files = {}
```
ladinesa: not sure why this should return a dictionary, can't this simply return a flat list of files ordered by priority? Maybe also some docstrings and a TODO maybe for future optimization?

atomisticparsers/gromacs/parser.py
```diff
@@ -1629,6 +1694,53 @@ def parse_input(self):
...
+        # Check if an MDAnalysis universe can be created from the mainfile and the current trajectory file
+        def matches_mainfile(file_path):
+            # set the trajectory file to the current file_path, the traj_parser attempts to create a universe
+            try:
+                self.traj_parser.auxilliary_files = [file_path]
+            except Exception as e:
+                # If the trajectory file does not match the topology, the trajectory file remains None
+                self.logger.warning(
+                    'Trajectory file does not match topology: {}, {}'.format(
+                        file_path, e
```
ladinesa: remove variables in log. add them as data if necessary.

ladinesa:
sorry just have a couple more suggestions then you can merge.

atomisticparsers/gromacs/parser.py
Bernadette-Mohr: We are currently assigning three priorities by filename to all files with a certain format. We want to consider the lowest priority only if no match can be found for any format within the two higher priorities. A flat list would remove this aspect.

ladinesa: but as far as i understand in l1730-1736, you are still concatenating them in a list and if the file is somehow readable you simply ignore the others in the list.

ladinesa: the other uses are, you only read the `[0]`, so in this case you can just add it as a kwarg, say a list of priorities to return?

ladinesa: anyway if this does not make sense, you can change the return to a tuple[list], a dict is really not the appropriate i would say.

atomisticparsers/gromacs/parser.py
Bernadette-Mohr: Done. The error message should have been logged by `MDAnalysisParser.universe` already. I am adding `file_path` as an `extra` right now, maybe which file didn't match could be valuable information for the user?

ladinesa: yes this could be helpful. the warning loops over all files, would be nicer actually to put it outside. but your call.

atomisticparsers/gromacs/parser.py
Bernadette-Mohr: > the other uses are, you only read the `[0]`, so in this case you can just add it as a kwarg, say a list of priorities to return?

I've extended `get_all_gromacs_files` by a `return_all` flag. By default, it returns a list with the highest-priority match it finds; if no match is found, `get_all_gromacs_files` returns an empty list. Only if the flag is set does it return all options in a tuple of three lists. 
I know functions with two different return types are usually not advised. Is this approach problematic w/r/t, e.g., type checking?

> anyway if this does not make sense, you can change the return to a tuple[list], a dict is really not the appropriate i would say.

Agreed. A tuple of three lists seems to be a more performant option.

atomisticparsers/gromacs/parser.py
```diff
@@ -810,19 +814,81 @@ def get_gromacs_file(self, ext):
...
+        """
+        Tries to find all the Gromacs output files with the given extension.
+        return_all: if True, return all files with the given extension, sorted by basename similarity
+        and uniqueness. If False, return the first file that matches the basename of the main file.
+        Returns:
+            - tuple: A tuple of two lists, the first list contains the files with at least partially
+            the same basename as the main file, the second list contains the files with different
+            basenames, sorted by uniqueness. (return_all == True)
+            - str: the first file that matches the basename of the main file. (return_all == False)
+            - None: if no file with the given extension is found.
```
JFRudzinski: returns [] if no file with the extension if found I think?

atomisticparsers/gromacs/parser.py
```diff
@@ -1629,6 +1699,57 @@ def parse_input(self):
...
+        """
+        traj_priority_list = ['trr', 'xtc', 'pdb', 'gro']
+
+        # Check if an MDAnalysis universe can be created from the mainfile and the current trajectory file
+        failed_files = []
+
+        def matches_mainfile(file_path):
+            # set the trajectory file to the current file_path, the traj_parser attempts to create a universe
+            try:
+                self.traj_parser.auxilliary_files = [file_path]
```
JFRudzinski: Does setting only the auxilliary file create the universe? I guess you know this I just haven't looked in a long time, and would have thought you need to call self.traj_parser.universe?

atomisticparsers/gromacs/parser.py
```diff
@@ -1629,6 +1699,57 @@ def parse_input(self):
...
+            except Exception as e:
+                # If the trajectory file does not match the topology, the trajectory file remains None
+                failed_files.append(file_path)
```
JFRudzinski, line 1706: is `failed_files` actually being saved? It's defined outside this function

atomisticparsers/gromacs/parser.py
```diff
@@ -1629,6 +1699,57 @@ def parse_input(self):
         )
         sec_control_parameters.m_set(quantity_def, input_parameters)
 
+    def find_trajectory_file(self) -> Union[str, None]:
```
JFRudzinski: Can you create a few tests which directly assess this function.

I am thinking you do:
test_parser = GromacsParser()
test_parser._gromacs_files = ["...", "..."] # set these by hand to manufacture different cases
result = test_parser.find_trajectory_file()
assert result...

JFRudzinski (commented):
@Bernadette-Mohr This is looking very nice and extensive!

Sorry for the possibly basic questions, it's just not been a while since I have looked at this, so I want to make sure that I am understanding.

My major concern is about testing this to make sure that it is functioning properly in various cases. I give one idea towards a unit-type test to address a bunch of different cases that you can think up.

I would also like to know that it is working reliably for existing cases. Beyond the tests within the parser, have you tried it on any Gromacs data on the repo? I would ask that you collect a variety of distinct Gromacs simulations, e.g., from the available datasets from myself and from Tristan, and try running at least one entry parsing from each to double check that everything is working as expected.

atomisticparsers/gromacs/parser.py
Bernadette-Mohr: You're right! I forgot to update the comments. Fixed it now.

atomisticparsers/gromacs/parser.py
Bernadette-Mohr: Yes, it does: `self.traj_parser` is an instance of `GromacsMDAnalysisParser`, which automatically attempts to create the universe and returns an error if that fails.

atomisticparsers/gromacs/parser.py
Bernadette-Mohr: So far, it is only logged as an `extra` if trajectories with the right extensions were found, but none matched the given topology file. This could be helpful information for the user in the future, if something goes wrong with their upload preparation and the file they intended to contain the correct trajectory is not accepted.
I moved the logging out of `matches_mainfile` to not unnecessarily clutter the NOMAD stdout.

atomisticparsers/gromacs/parser.py
JFRudzinski: Please remove the old function now :+1: , and then resolve this

atomisticparsers/gromacs/parser.py
JFRudzinski: The small overhead is fine for now, maybe we can assess the performance of everything in the new parser, but let's not dwell upon it now (you can resolve this one)

JFRudzinski (commented):
@Bernadette-Mohr For me this looks good now :raised_hands: 

TODOs:
1. Test a variety of repo data as previously discussed
2. Add the unit tests in the placeholder function that you added
3. Clear the MR with Alvin
4. Merge (and after I will make a release)

atomisticparsers/gromacs/parser.py
Bernadette-Mohr: The warning is moved to the end of `find_trajectory_file` now, it is only executed if no matching trajectory has been found, and only once.

atomisticparsers/gromacs/parser.py
```diff
@@ -363,10 +363,14 @@ def parse(self, key):
...
 
     def keys(self):
+        if self.fileedr is None:
+            return []
         return list(self.fileedr.keys())
 
     @property
     def length(self):
+        if self.fileedr is None:
+            return None
```
ladinesa: return 0 here probably

atomisticparsers/gromacs/parser.py
```diff
@@ -772,57 +776,87 @@ def get_mdp_file(self):
...
         # TODO read also from ene
-        edr_file = self.get_gromacs_file('edr')
-        self.energy_parser.mainfile = edr_file
-
-        # get it from edr file
-        if self.energy_parser.keys():
+        # Make sure *.edr file is part of upload before attempting to parse it.
+        edr_file = next(iter(self.get_gromacs_files('edr')), None)
+        if edr_file:
+            self.energy_parser.keys()
```
ladinesa: why necessary to access keys?

atomisticparsers/gromacs/parser.py
```diff
@@ -772,57 +776,87 @@ def get_mdp_file(self):
...
+    def get_gromacs_files(
+        self, ext, return_all=False
+    ) -> Union[Tuple[List[str], List[str], List[str]], List[str]]:
+        """
+        Tries to find all the Gromacs output files with the given extension.
+        return_all: If True, return all files with the given extension, sorted
+        by basename similarity and uniqueness. If False, return the first file
+        that matches the basename of the main file.
+        Returns:
+        - tuple: return_all == True. A tuple of two lists.
```
ladinesa: tuple of three lists?

atomisticparsers/gromacs/parser.py
Bernadette-Mohr: Good point, return type is changed.

atomisticparsers/gromacs/parser.py
```diff
@@ -1690,30 +1794,9 @@ def write_to_archive(self):
...
-        self.traj_parser.auxilliary_files = [trajectory_file]
-
-        # check to see if the trr file can be read properly (and has positions), otherwise try xtc file instead
-        positions = None
-        if (universe := self.traj_parser.universe) is not None:
-            atoms = getattr(universe, 'atoms', None)
-            positions = getattr(atoms, 'positions', None)
-        if positions is None:
-            self.traj_parser.auxilliary_files = [xtc_file] if xtc_file else [trr_file]
+        # Trajectory file is passed to MDAnalysisParser parser under the hood
```
ladinesa: I do not prefer this, I suggest have, `parse_trajectory` where you assign the traj file from `find_trajectory_file`

atomisticparsers/gromacs/parser.py
Bernadette-Mohr: I think this was left over from an early attempt at handling errors caused by missing or corrupted edr files. It is not needed anymore, I removed it.

atomisticparsers/gromacs/parser.py
Bernadette-Mohr: Yepp, fixed.

atomisticparsers/gromacs/parser.py
ladinesa: Do you think it is a good idea to loop over the match files and see which edr file can be read, pretty much like what you are doing now for the traj files? this also goes to other file types.

atomisticparsers/gromacs/parser.py
```diff
@@ -1629,6 +1663,78 @@ def parse_input(self):
...
+        # sorted by extension priority (priority_list) and filename.
+        # Select matching trajectory with highest priority.
+        def try_priority_group(traj_priority_group) -> bool:
+            contains_files: List[str] = []
+            fallback_files: List[str] = []
+            for file_ext in traj_priority_group:
+                traj_files_tup = self.get_gromacs_files(file_ext, return_all=True)
+                if traj_files_tup:
+                    for file_path in traj_files_tup[0]:
+                        if matches_mainfile(file_path):
```
ladinesa: i suggest you do not do `matches_mainfile` in a loop, i suggest you simply get a list of all files arranged in priority for all groups and extensions, searching files is not really expensive, this way code is cleaner, i.e. no explosion of `return` statements, you can even put some of the logic in `get_gromacs_files` such that it accepts a also list of ext.

Bernadette-Mohr: I have restructured the logic of `find_trajectory_file` a bit, there is only one loop over the candidate files left now. I am not positive moving some of the prioritizing logic into `get_gromacs_files` would make sense, since this function already doubles for only selecting a single file with a single extension for the other use cases outside of `find_trajectory_file?

atomisticparsers/gromacs/parser.py
Bernadette-Mohr: I think the main issue here is that not every GROMACS simulation writes out an edr file in the first place, the original code used to rais an error if there was no edr file in the upload.

PR #154, opened 2025-03-26
Remove nptyping

ladinesa:


pyproject.toml
```diff
@@ -20,6 +20,9 @@ dependencies = [
     "panedr>=0.2",
     "scipy>=1.7.1",
     "pydantic>=1.10.8",
+    "ase~=3.22.0",
+    "numpy>=1.22.4,<2",
```
blueraft: Can this be relaxed to `numpy>=1.22.4`? `nomad-lab` can handle the upper bound restriction here.

PR #155, opened 2025-05-02
Add single point workflow

ladinesa:


schumannj:
@ladinesa I tested the branch and it works for my single point trajectory files. Thanks!

schumannj (approved):

PR #159, opened 2025-08-01
Debug lammps leo

leojbeck:
Goals:
- Editing parser regex to work on a few more version formats and thermodynamic data formats (both from log file)
- Extend to atom style for the dump files
- Fix pbc_cell logic to allow for triclinic cells

To Dos:

- [ ] Clean comments and print statements
- [x] Fix ruff formatting errors
- [ ] Fix tests
- [ ] Get review and address comments
- [ ]


JFRudzinski:
@leojbeck I fixed the ruff errors for you, they had nothing to do with your code, just some new rules from ruff affecting old code

JFRudzinski:
@ladinesa can you identify the source of these errors?

I reproduce locally with 1. this branch, 2. this branch rebased to develop...but if I just checkout develop I don't have these errors. And I can't see anything relevant that is changed in this PR :thinking:

ladinesa:
The module-level import for MatchingParserInterface broke this, It will lead to cyclic import, pushed fix now.

PR #167, opened 2025-12-11
workflow scheema stuff

JFRudzinski:
@ndaelman-hu I made a separate PR to point this out:

There were some missing quantities in the Gromacs features yaml, corresponding to quantities that are automatically populated by the schema normalization (in this case in MD.results). The question is how we deal with such things. In this branch I asked the model to look at the workflow schema and adjust the features and instructions accordingly. However, I am not sure if what it added to the instructions is out of scope.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -74,7 +74,7 @@ Describes the computational methodology used.
     - `atom_indices` - Atom indices
     - `n_interactions` - Number of interactions of this type
   - Force fields supported in this repository: AMBER, GROMOS, OPLS, COMPASS (via specific parsers or force field parameters)
-  
+
 #### Tight-Binding Methods (Semi-empirical)
 - `TB` - Tight-binding method container
   - `name` - Method name: "DFTB", "xTB"
@@ -311,10 +311,10 @@ runschema_capabilities:
     # Time-dependent properties
     - time_physical  # for MD trajectories
     - step  # time step number
-    
+
     # Vibrational properties
     - vibrational_frequencies  # if phonons/vibrations calculated
-    
+
     # Constraints
     - constraint  # geometric constraints on atoms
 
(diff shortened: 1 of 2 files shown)


.github/copilot-instructions.md
```diff
@@ -473,6 +473,99 @@ elif self.is_elastic_calculation():
...
+```python
+# DON'T calculate RDF in parser - normalizer does this
+workflow.results.radial_distribution_functions = compute_rdf(trajectory)
+
+# DON'T calculate MSD in parser - normalizer does this
+workflow.results.mean_squared_displacements = compute_msd(trajectory)
+
+# DON'T calculate radius of gyration in parser - normalizer does this
+workflow.results.radius_of_gyration = compute_rg(trajectory)
+```
```
ndaelman-hu: What is this for? To review code quality?

.github/copilot-instructions.md
```diff
@@ -473,6 +473,99 @@ elif self.is_elastic_calculation():
...
+
+### What Parsers Should Populate
+
+Parsers should focus on:
+1. **Raw trajectory data**: Store system snapshots in `run.system[]` with positions, velocities, forces
+2. **Per-frame thermodynamics**: Store temperature, pressure, volume, energy in `calculation.thermodynamics`
+3. **Workflow method parameters**: ensemble type, thermostat, barostat, timestep, etc.
+4. **Basic workflow results**: `n_steps`, `trajectory` reference, `finished_normally`
+
+The normalizers will automatically:
```
ndaelman-hu, line 719: This is neat, but what would be the use case?

.github/copilot-instructions.md
```diff
@@ -473,6 +473,99 @@ elif self.is_elastic_calculation():
...
+# DON'T calculate MSD in parser - normalizer does this
+workflow.results.mean_squared_displacements = compute_msd(trajectory)
+
+# DON'T calculate radius of gyration in parser - normalizer does this
+workflow.results.radius_of_gyration = compute_rg(trajectory)
+```
+
+### Example: What TO do
+```python
+# DO store trajectory snapshots
```
ndaelman-hu, line 739: Could we see how effective these "DO" statements are vs bold-face "Important"?

ndaelman-hu (commented):
Some interesting additions. Pls also take a look to what I added for the [electronic parsers](https://github.com/nomad-coe/electronic-parsers/pull/321) and the [ultimate stress test](https://github.com/nomad-coe/electronic-parsers/pull/324).

I would aim to execute the latter as a final test, for sure.
No opinion on the schema-specific updates, though.

.github/copilot-instructions.md
JFRudzinski: Yeah, I mean I guess it is just clarifying what is expected to be found in the parsers because I asked it to write down which features are to be added always based on the normalizer ... but I think this maybe is going too far, or out of scope for our particular usage here right?

.github/copilot-instructions.md
JFRudzinski: I needed to add this so that the corresponding features eg calculation of rdf, msd, etc would be added to the Gromacs features.yml even though they do not appear directly in the parser code. Should I not do this? Should I only documents quantities that are parsed directly?

.github/copilot-instructions.md
JFRudzinski: I guess we might remove this section as mentioned above, but yes if there is another case of this usage we can test the alternative

